PR #7, opened 2019-12-11
Feature / License

bryantbiggs:
- Add Apache 2.0 License


michaeljguarino:
I'm not sure that I intend to use an apache license for this long term, but for now I think it's ok

PR #156, opened 2022-01-11
Init add equinix metal

DavidSpek:


michaeljguarino (commented):
Same as in the console, we should probably find a better image that'll work as an icon throughout the app.

DavidSpek:
@michaeljguarino I've changed the icon

michaeljguarino (approved):

PR #172, opened 2022-02-21
TF Provider Template API's

ycryptx:
## Summary
Adds API's that return terraform scaffolds of our supported providers. This is so that the CLI is not hard coded with those scaffolding, to allow us to more seamlessly update those terraform definitions as needed.

## Test Plan
Added unit tests that cover the 2 new gqp api's as well as tests covering the underlying logic

## Checklist
- [X] I have added tests to cover my changes.

### DIFF
--- a/apps/core/lib/core/schema/provider_scaffold.ex
+++ b/apps/core/lib/core/schema/provider_scaffold.ex
@@ -1,0 +1,18 @@
+defmodule Core.Schema.ProviderScaffold do
+  defstruct [:name, :content]
+  @providers ~w(aws gcp azure equinix)a
+
+  def available(), do: @providers
+
+  def new(name) when name in @providers do
+    %__MODULE__{
+      name: name,
+      content: provider_file(name)
+    }
+  end
+
+  defp provider_file(name) do
+    Path.join([:code.priv_dir(:core), "scaffolds/terraform/providers", "#{name}.eex"])
+    |> File.read!
+  end
+end
--- a/apps/core/lib/core/services/scaffolds.ex
+++ b/apps/core/lib/core/services/scaffolds.ex
@@ -43,6 +43,10 @@ defmodule Core.Services.Scaffolds do
     end)
   end
 
+  def available_providers(), do: Core.Schema.ProviderScaffold.available()
+
+  def provider(name), do: Core.Schema.ProviderScaffold.new(name)
+
   defp eval(file, application, provider, ctx) do
     scaffold_file(file)
     |> eval_file(
--- a/apps/core/priv/scaffolds/terraform/providers/gcp.eex
+++ b/apps/core/priv/scaffolds/terraform/providers/gcp.eex
@@ -1,0 +1,49 @@
+terraform {
+  backend "gcs" {
+    bucket = {{ .Values.Bucket | quote }}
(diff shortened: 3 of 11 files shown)


apps/core/lib/core/schema/provider_scaffold.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Core.Schema.ProviderScaffold do
+  defstruct [:name, :content]
+  @providers ~w(aws gcp azure equinix)a
+
+  def available() do
```
michaeljguarino, line 5: small style nit, but you can make these one-line functions a bit more concise with:

```elixir
def available(), do: @available
```

apps/core/lib/core/services/scaffolds.ex
```diff
@@ -43,6 +43,14 @@ defmodule Core.Services.Scaffolds do
     end)
   end
 
+  def available_providers() do
```
michaeljguarino, line 46: similar, use one-line syntax

apps/core/lib/core/services/scaffolds.ex
```diff
@@ -43,6 +43,14 @@ defmodule Core.Services.Scaffolds do
...
+    Core.Schema.ProviderScaffold.available()
+  end
+
+  def provider(name) do
```
michaeljguarino, line 50: use one-line syntax

apps/core/priv/scaffolds/terraform/providers/gcp.eex
```diff
@@ -0,0 +1,154 @@
...
+
+provider "kubernetes" {
+  host = data.google_container_cluster.cluster.endpoint
+  cluster_ca_certificate = base64decode(data.google_container_cluster.cluster.master_auth.0.cluster_ca_certificate)
+  token = data.google_client_config.current.access_token
+}
+{{ end }}
+`
+
+const awsBackendTemplate = `terraform {
```
michaeljguarino, line 52: I'm pretty sure this file is wrong if this is in here?

apps/graphql/lib/graphql/schema/scaffold.ex
```diff
@@ -0,0 +1,22 @@
+defmodule GraphQl.Schema.Scaffold do
+  use GraphQl.Schema.Base
+
+  alias GraphQl.Resolvers.Scaffolds
+
+  object :tfprovider do
```
michaeljguarino, line 6: I think making this type named `terraform_provider` or something a bit less abbreviated might be ideal so that external devs wouldn't be confused

apps/core/test/services/scaffolds_test.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Core.Services.ScaffoldsTest do
+  use Core.SchemaCase, async: true
+  alias Core.Services.Scaffolds
+
+  describe "#tf providers" do
```
michaeljguarino, line 5: Could we add a test that validates that all providers are implemented, maybe just doing:

```elixir
Scaffolds.available_providers()
|> Enum.map(&Scaffolds.provider/1)
```

and ensuring nothing throws?

michaeljguarino (approved):

PR #185, opened 2022-03-16
GCP CLI use Region so it is consistent with other providers

DavidSpek:
## Summary
Accompanying change for https://github.com/pluralsh/plural-cli/pull/30. The template has been updated to reflect the CLI change where the GCP Region is used rather than the Zone.

## Checklist

### DIFF
--- a/apps/core/priv/scaffolds/terraform/providers/gcp.eex
+++ b/apps/core/priv/scaffolds/terraform/providers/gcp.eex
@@ -16,15 +16,9 @@ terraform {
   }
 }
 
-locals {
-  gcp_location  = {{ .Values.Location | quote }}
-  gcp_location_parts = split("-", local.gcp_location)
-  gcp_region         = "${local.gcp_location_parts[0]}-${local.gcp_location_parts[1]}"
-}
-
 provider "google" {
   project = {{ .Values.Project | quote }}
-  region  = local.gcp_region
+  region  = {{ .Values.Region | quote }}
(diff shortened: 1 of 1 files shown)


michaeljguarino (approved):

apps/core/priv/scaffolds/terraform/providers/gcp.eex
```diff
@@ -16,15 +16,9 @@ terraform {
...
-locals {
-  gcp_location  = {{ .Values.Location | quote }}
-  gcp_location_parts = split("-", local.gcp_location)
-  gcp_region         = "${local.gcp_location_parts[0]}-${local.gcp_location_parts[1]}"
-}
-
 provider "google" {
   project = {{ .Values.Project | quote }}
-  region  = local.gcp_region
+  region  = {{ .Values.Region | quote }}
```
michaeljguarino, line 21: Actually wait, since this value is not provided in old cli versions, there will be users who it will break for.  We should default it to `.Values.Location` if `.Values.Region` is not present

michaeljguarino (changes requested):

apps/core/priv/scaffolds/terraform/providers/gcp.eex
DavidSpek: Yeah that's a good idea.

michaeljguarino (approved):

PR #188, opened 2022-03-16
Add provider versioning

### LINKED ISSUE #186: Provider Scaffolds api should support versioning
## Use Case
It's too difficult to develop new versions of providers using the current set up, with a versioned approach we can roll things out more iteratively for specific packages

## Ideas of Implementation
* store each version in a directory in priv
* add arg in graphql

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

michaeljguarino:
## Summary

It's still tricky to upgrade provider templates since the blast radius is so large.  Adding a version
system and allowing specific packages to specify the version they want can theoretically mitigate this.

## Test Plan
unit tests pass

## Checklist

Closes #186

### DIFF
--- a/apps/core/priv/scaffolds/terraform/providers/1.0/azure.eex
+++ b/apps/core/priv/scaffolds/terraform/providers/1.0/azure.eex
@@ -41,47 +41,4 @@ provider "kubernetes" {
   client_key             = base64decode(data.azurerm_kubernetes_cluster.cluster.kube_config[0].client_key)
   cluster_ca_certificate = base64decode(data.azurerm_kubernetes_cluster.cluster.kube_config[0].cluster_ca_certificate)
 }
-{{ end }}terraform {
-  backend "azurerm" {
-    storage_account_name = {{ .Values.Context.StorageAccount | quote }}
-    resource_group_name = {{ .Values.ResourceGroup | quote }}
-    container_name = {{ .Values.Bucket | quote }}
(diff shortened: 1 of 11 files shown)


apps/core/priv/scaffolds/terraform/providers/1.0/azure.eex
```diff
@@ -0,0 +1,87 @@
...
+  resource_group_name = {{ .Values.ResourceGroup | quote }}
+}
+
+provider "kubernetes" {
+  host                   = data.azurerm_kubernetes_cluster.cluster.kube_config[0].host
+  client_certificate     = base64decode(data.azurerm_kubernetes_cluster.cluster.kube_config[0].client_certificate)
+  client_key             = base64decode(data.azurerm_kubernetes_cluster.cluster.kube_config[0].client_key)
+  cluster_ca_certificate = base64decode(data.azurerm_kubernetes_cluster.cluster.kube_config[0].cluster_ca_certificate)
+}
+{{ end }}terraform {
```
DavidSpek, line 44: It looks like this was copied twice. I believe this was already the case with the first iteration of managing the provider template through the API since I noticed it while deploying on Azure.

michaeljguarino: yeah you're right, good catch

PR #190, opened 2022-03-22
Add Kind as a provider in Plural.

DavidSpek:
## Summary
Add Kind as a provider in Plural.

## Checklist


apps/core/priv/scaffolds/terraform/providers/kind.eex
```diff
@@ -0,0 +1,25 @@
...
+
+  required_providers {
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = "~> 2.8.0"
+    }
+  }
+}
+
+{{- if .Values.ClusterCreated }}
```
michaeljguarino, line 14: shouldn't this be the inverse?

DavidSpek: Copied from the Equinix setup. `ClusterCreated` seems to be if the cluster is being deployed during that step. The wording is a bit confusing though, I remembered having it reversed at first with Equinix.

PR #193, opened 2022-03-30
Add "Log back as previous user" feature

### LINKED ISSUE #180: Return to Previous User after assuming a service account identity
## Use Case
Currently once you've assumed a service account, the only way to return to your primary user is by logging out and logging back in.  We should keep track of your base user and have a way to always easily return as needed.

## Ideas of Implementation
Probably store base user jwt somewhere else in local storage, along with a copy of the user record for things like remembering name/email/avatar.

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

dherault:
## Summary
Fixes #180 
image

### DIFF
--- a/www/src/components/accounts/ServiceAccounts.js
+++ b/www/src/components/accounts/ServiceAccounts.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react'
+import React, { useCallback, useState, useContext } from 'react'
 import { Box, Layer, TextInput } from 'grommet'
 import { useMutation } from 'react-apollo'
 import { Edit } from 'grommet-icons'
@@ -9,13 +9,14 @@ import { UserRow } from './User'
 import { extendConnection, removeConnection, updateCache } from '../../utils/graphql'
 import { useQuery } from 'react-apollo'
 import { GqlError } from '../utils/Alert'
-import { setToken } from '../../helpers/authentication'
+import { setToken, setPreviousUserData, fetchToken } from '../../helpers/authentication'
 import { SearchIcon } from './utils'
 import { SectionContentContainer, SectionPortal } from '../Explore'
 import { INPUT_WIDTH } from './constants'
 import { LoopingLogo } from '../utils/AnimatedLogo'
 import { Icon } from './Group'
 import { DeleteUser } from '../users/DeleteUser'
+import { CurrentUserContext } from '../login/CurrentUser'
 
 function ServiceAccount({user, next, update}) {
   const [open, setOpen] = useState(false)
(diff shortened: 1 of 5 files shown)


www/src/components/accounts/ServiceAccounts.js
```diff
@@ -27,12 +28,21 @@ function ServiceAccount({user, next, update}) {
       window.location = '/'
     }
   })
+  const me = useContext(CurrentUserContext)
+
+  function handleImpressonateClick() {
```
michaeljguarino, line 33: rename to `handleImpersonateClick`.  Also I think this should probably be a `useCallback` hook accepting mutation as a dependency, although I suspect it wouldn't actually matter

michaeljguarino (commented):
I feel like this call-to-action is slightly unintuitive.  Could it potentially be another element above or below logout with similar styling (you can probably use that switch icon if you'd like).

www/src/helpers/authentication.js
```diff
@@ -10,4 +10,17 @@ export function fetchToken() {
 
 export function setToken(token) {
   localStorage.setItem(AUTH_TOKEN, token)
-}
\ No newline at end of file
+}
+
+export function setPreviousUserData(userData) {
+  localStorage.setItem(AUTH_PREVIOUS_USER_DATA, JSON.stringify(userData))
```
michaeljguarino, line 16: Does this have to be stringified?  I feel like local storage should be able to just store the raw user.  Alternatively, we could just have this store only the jwt as that's all that's necessary.

www/src/components/accounts/ServiceAccounts.js
dherault: Good catch, it's my GitHub Copilot's mistake :)
No need for a useCallback here imo.

www/src/helpers/authentication.js
dherault: Yes localStorage can only store string key/value pairs.

dherault:
image

www/src/helpers/authentication.js
michaeljguarino: in that case, perhaps we should just store the jwt here to avoid the annoying json serialization logic, esp since I believe it's all that's need to restore the previous user.

michaeljguarino: Oh wait, I see, you're trying to keep name and avatar and all that, that makes sense

www/src/helpers/authentication.js
```diff
@@ -10,4 +10,17 @@ export function fetchToken() {
...
+}
+
+export function getPreviousUserData() {
+  try {
```
michaeljguarino, line 20: I wonder if we can avoid the try catch here then, perhaps:

```js
const stored = localStorage.getItem(AUTH_PREVIOUS_USER_DATA)
return stored && JSON.parse(stored)
```

michaeljguarino: If that's actually best practice, Let's leave as-is

michaeljguarino (approved):

PR #194, opened 2022-03-30
Lint application

dherault:
## Summary
After one hour of investigation, I found out that I should not touch the hook deps blindly :)

This fixes all lint errors.
image

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -91,7 +91,13 @@
     "eject": "react-scripts eject"
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": "react-app",
+    "rules": {
+      "default-case": "off",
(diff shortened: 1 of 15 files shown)


www/package.json
```diff
@@ -91,7 +91,13 @@
     "eject": "react-scripts eject"
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": "react-app",
+    "rules": {
+      "default-case": "off",
+      "no-mixed-operators": "off",
+      "react-hooks/exhaustive-deps": "off",
```
michaeljguarino, line 98: I didn't know you could do this!  very good fix

michaeljguarino (approved):

PR #209, opened 2022-04-18
Implement new sidebar

dherault:
- Use new sidebar 
- Lint
- Fix imports
- Remove unused files (still available in the history)


michaeljguarino (approved):

www/src/components/Breadcrumbs.js
```diff
@@ -75,9 +77,10 @@ export function Breadcrumbs() {
 
 export default function BreadcrumbProvider({ children }) {
   const [breadcrumbs, setBreadcrumbs] = useState([])
+  const value = useMemo(() => ({ breadcrumbs, setBreadcrumbs }), [breadcrumbs])
```
dherault, line 80: These useMemo prevent a lot of useless rerendering of the application. It makes it way faster hence.

michaeljguarino: very helpful

michaeljguarino (approved):

PR #210, opened 2022-04-20
Implement top bar

dherault:
image

### DIFF
--- a/www/src/components/repos/SearchRepositories.js
+++ b/www/src/components/repos/SearchRepositories.js
@@ -2,9 +2,8 @@ import { useState } from 'react'
 import styled from 'styled-components'
 import { useApolloClient } from '@apollo/client'
 import { useNavigate } from 'react-router-dom'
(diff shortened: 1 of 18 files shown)


dherault:
Please wait for https://github.com/pluralsh/design-system/pull/22 to be merged before reviewing.

www/src/components/repos/SearchRepositories.js
```diff
@@ -49,6 +41,7 @@ export default function SearchRepositories() {
           type="search"
           value={value}
           name="search"
+          icon={<MagnifyingGlassIcon color="text-weak" />}
```
michaeljguarino, line 44: One thing I was going to mention to the designers, but the backslash search icon is a bit nicer for us than the plain glass icon, and I've seen a lot of dev-y products start using it.  I think we should keep it.

michaeljguarino (commented):
Minor fix that's not your fault, just a bit of lag on changing the design based on some of my feedback

PR #221, opened 2022-05-03
Design fixes 4

dherault:
Do not merge yet,  needs a newer version of pluralsh-design-system


www/src/components/Socials.js
```diff
@@ -1,39 +1,33 @@
...
 export function SocialLinks() {
   return (
-    <Box
-      flex={false}
-      margin={{ left: '16px' }}
-      direction="row"
-      align="center"
+    <Div
+      ml={1}
+      xflex="x4"
```
michaeljguarino, line 29: We should find a more self-documenting convention for these as well.

michaeljguarino (approved):
Approving to unblock, but we should think more about self-documenting props overall.  It's going to be a hurdle for new devs to understand the code as-is.

PR #244, opened 2022-06-01
Complete Marketplace Design

dherault:
## Summary

Resolves ENG-118

Finalizes the layout for the "Marketplace" and "Installed" scenes.

image
## Checklist


www/src/components/marketplace/MarketplaceRepositories.js
```diff
@@ -71,6 +78,14 @@ function MarketplaceRepositories({ installed }) {
     })
     .filter(repository => installed ? repository.installation : true)
 
+  const fuse = new Fuse(sortedRepositories, searchOptions)
+
+  const resultRepositories = search ? fuse.search(search).map(({ item }) => item) : sortedRepositories
```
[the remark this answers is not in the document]
dherault: Yes, this hack works well for now.
In the future we might want to use the open-source version of Algolia 🇫🇷 !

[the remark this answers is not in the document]
michaeljguarino: no we should use the api's own search features.

www/src/components/ModalHeader.js
```diff
@@ -21,7 +21,7 @@ export function ModalHeader({ text, setOpen }) {
           justify="center"
           width="2rem"
           height="2rem"
-          hoverIndicator="background-light"
+          hoverIndicator="fill-one"
```
michaeljguarino, line 24: the name fill-one seems a little unclear to me, is there a way to make it a little more apparent what that color is acting as?

www/src/components/marketplace/Marketplace.js
```diff
@@ -53,32 +53,30 @@ function Marketplace({ installed }) {
...
+          transform={areFiltersOpen ? 'translateX(0)' : 'translateX(100%)'}
+          opacity={areFiltersOpen ? 1 : 0}
           flexShrink={0}
           position="sticky"
           top={0}
-          left={0}
-          width={areFiltersOpen ? 256 - 32 : 0}
-          height={`calc(100% - ${32}px)`}
+          right={0}
+          width={256 - 32}
```
michaeljguarino, line 73: is there a reason why we don't just make thse constants?

www/src/components/marketplace/Marketplace.js
```diff
@@ -53,32 +53,30 @@ function Marketplace({ installed }) {
...
+          height={`calc(100% - ${64 + 8}px)`}
```
michaeljguarino, line 74: same shouldn't this just be a const?

www/src/components/marketplace/MarketplaceSidebar.js
```diff
@@ -59,96 +52,127 @@ function MarketplaceSidebar(props) {
...
+    )
   }
 
-  const sortedCategories = categoriesData.categories.slice().sort((a, b) => a.category.localeCompare(b.category))
-  const sortedPublishers = ['Plural']
-  const sortedTags = tags.slice().sort((a, b) => a.tag.localeCompare(b.tag)).filter((x, i) => i < nDisplayedTags)
+  function renderTags() {
+    const sortedTags = tags.slice().sort((a, b) => a.name.localeCompare(b.name))
+    const fuse = new Fuse(sortedTags, searchOptions)
+    const resultTags = search ? fuse.search(search).map(({ item }) => item) : sortedTags.filter((x, i) => i < nDisplayedTags)
```
michaeljguarino, line 121: we're definitely not going to be able to load all tags in one fetch in relatively short order, so this should really be moved to utilizing the gql apis properly

www/src/components/ModalHeader.js
dherault: Ask Nick, that's his responsibility

www/src/components/marketplace/Marketplace.js
dherault: I will

www/src/components/marketplace/MarketplaceSidebar.js
dherault: I'm waiting on some green go from the backend team then.

michaeljguarino: what do you need api wise?  I'm pretty sure we do have the capability to filter tags

www/src/components/ModalHeader.js
michaeljguarino: We can bring it up with nick too, but for codebase cleanliness, I do think we should have a mapping of semantic names that correspond to his more systematic ones.

www/src/components/marketplace/MarketplaceSidebar.js
michaeljguarino: yep, you can use the `q` argument for the tags query to filter by a search string

www/src/components/marketplace/MarketplaceRepositories.js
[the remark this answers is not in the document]
michaeljguarino: You should now have the `categories`, `publishers` and `tags` fields to do multi-value filtering for all of these via the api.  We should definitely strongly prefer that to in-browser filtering

www/src/components/marketplace/MarketplaceRepositories.js
```diff
@@ -61,16 +64,33 @@ function MarketplaceRepositories({ installed }) {
 
   const sortedRepositories = repositories.slice()
     .sort((a, b) => a.name.localeCompare(b.name))
-    .filter(repository => categories.length ? categories.includes(repository.category) : true)
+    .filter(repository => categories.length ? categories.includes(repository.category.toLowerCase()) : true)
```
michaeljguarino, line 67: sorting should also be delegated to the api

dherault: I've added a ticket for you to do it.

dogmar:
Scrollbar should appear on right edge of browser window. Otherwise, looks great.
Screen Shot 2022-06-07 at 11 05 04 AM

dherault:
Will work on it.

dogmar:
Marketplace scrollbar is good. I should have caught this earlier, but Filters sidebar is overflowing horizontally and showing a horizontal scrollbar.

Screen Shot 2022-06-07 at 1 12 42 PM

michaeljguarino (approved):
This looks good, make sure I have a ticket to fix up the api usage and let's press on.

PR #286, opened 2022-06-30
Fix sidebar tooltip styling - Closes ENG-192

dogmar:


www/src/components/layout/Sidebar.js
```diff
@@ -138,14 +139,16 @@ ref
   }
 
   function wrapTooltip(node) {
-    if (!(collapsed && tooltip)) return node
+    if (!(tooltip)) return node
```
michaeljguarino, line 142: the inner parens are unnecessary here I think

dogmar:
> 

The animations are working now on my end. Are you still seeing issues @dherault?

> This breaks the animation transitions on the sidebar. I'm pro merging anyway.

PR #290, opened 2022-07-01
Add cloud shell Cheatsheet, complete cloud shell layout

dherault:
## Summary

Completes the layout of the Cloud Shell

## Checklist


www/src/components/shell/TerminalSidebar.js
```diff
@@ -422,4 +439,121 @@ function Step3() {
...
+      CheatsheetDebugging
+    </>
+  )
+}
+
+function CheatsheetWorkspace() {
+  return (
+    <>
+      CheatsheetWorkspace
+    </>
```
dherault, line 555: I need the data for these

michaeljguarino (approved):

www/src/components/shell/TerminalSidebar.js
michaeljguarino: I'm not actually sure which part of the mocks this refers to.  You should be able to get all the needed data from the query for the cloud shell instance itself

PR #293, opened 2022-07-02
Marketplace Fixes – ENG-214

dogmar:
Fixes issues in ENG-214, but Install Button icon is still not updated, as that requires a change in honorable.


michaeljguarino:
can you make this PR against www2?

PR #295, opened 2022-07-04
Add breadcrumbs to the marketplace repositories - ENG-204

floreks:
Restored breadcrumbs for the marketplace repositories.

Normal
image

Hovered
image

### Questions
1. Should we still consider `disabled` state for breadcrumbs? Design does not cover that. Is it used?
2. What about colors for different themes (dark/light)? Should I ignore that for now and simply reuse colors from the figma?
3. Should the repository application name be always capitalized?

### DIFF
--- a/www/src/components/Breadcrumbs.js
+++ b/www/src/components/Breadcrumbs.js
@@ -1,16 +1,17 @@
-import { createContext, useContext, useMemo, useState } from 'react'
-import { useNavigate } from 'react-router-dom'
-import { Anchor, Box, Text } from 'grommet'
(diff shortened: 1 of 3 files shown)


michaeljguarino:
1. I think we can ignore disabled state for now until designs call for it
2. We haven't built out light mode yet, but should mark it as a todo when we do. The theming should be flexible enough that the colors should switch seamlessly based on that setting
3. We can just use the repo name in the api (whihc is never capitalized i believe)

michaeljguarino (approved):

PR #305, opened 2022-07-05
ENG-227 - Hide the notification completely when on sidebar element not hovered

floreks:
## Summary

I guess this is a bit of a workaround. It might be the best to update the original `Tooltip` element that comes from the `honorable` library, so that we do not unknowingly fall into the same trap again when using the `Tooltip` element in other places.

## Test Plan

Tested locally

## Checklist

### DIFF
--- a/www/src/components/layout/Sidebar.js
+++ b/www/src/components/layout/Sidebar.js
@@ -113,6 +113,8 @@ function SidebarItemRef({
 },
 ref
 ) {
+  const [hovered, setHovered] = useState(false)
+
   function wrapLink(node) {
     if (!linkTo) return node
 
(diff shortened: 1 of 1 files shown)


dherault (approved):

www/src/components/layout/Sidebar.js
```diff
@@ -158,6 +161,8 @@ ref
 
   return wrapLink(wrapTooltip(
     <Flex
+      onMouseEnter={() => setHovered(true)}
```
michaeljguarino, line 164: onMouseEnter can sometimes be an unreliable event, I think you can actually use the `_hover` attribute to do this, right @dherault ?

michaeljguarino (approved):
we can merge this just to keep moving but might be worth thinking about a more robust solve

www/src/components/layout/Sidebar.js
dherault: _hover works only for css-related styling. Might work here.

PR #311, opened 2022-07-05
Properly redirect to onboarding from signup

michaeljguarino:
## Summary

Use our new onboarding state to differentiate oauth signups and mark onboarded from there

## Test Plan
test using `make web` and `mix test` for server changes

## Checklist

Fixes ENG-158


www/src/components/shell/CloudShell.js
```diff
@@ -100,6 +101,7 @@ function CloudShell() {
       childIsReady={ready}
     >
       <CreateRepositoryCard data={data} />
+      <OnboardingFooter />
```
maciaszczykm, line 104: I skipped it for cloud shell as I initially thought that onboarding will have different flow, where the side menu is not visible.

To avoid putting it in multiple places I'd move it from here and `OnboardingFlow.js` to `OnboardingWrapper.js` as it is being used in both places.

dherault (approved):

maciaszczykm (approved):

apps/graphql/lib/graphql/schema/repository.ex
```diff
@@ -218,6 +218,8 @@ defmodule GraphQl.Schema.Repository do
     field :auth_method,   non_null(:oidc_auth_method)
     field :configuration, :ouath_configuration, resolve: &GraphQl.Resolvers.OAuth.resolve_configuration/2
 
+    field :consent,  :consent_request
```
DavidSpek, line 221: @michaeljguarino This shouldn't be part of the OIDC provider, since it isn't related to the OIDC Provider setup. The `oauthConsent` query should return `Repository` as well as `consent_request`.

michaeljguarino: it's only filled in the oauth consent request, which actually returns an instance of the oidc provider I believe

floreks: The `oidcProvider` is part of the `installation` and i.e. when I was trying to log in to the console with a link that @DavidSpek shared with me, the `installation` object was empty for me. The structure seems to look like that.

```
oauthConsent {
      installation {
          oidcProvider {
              consent {
                  requestedScope
                  skip
              }
          }
      }
  }
```

PR #319, opened 2022-07-06
ENG-274 - Implement customized scroll bar

floreks:
## Summary

Updated the global scrollbar styling. 

@dherault @michaeljguarino do you know a better way to apply this kind of styling globally? This will not really work with theming since our theming configuration is inside the JS code.

Zrzut ekranu z 2022-07-06 15-25-23
Zrzut ekranu z 2022-07-06 15-25-50

## Test Plan

Locally

## Checklist

### DIFF
--- a/www/src/components/repos/Repositories.js
+++ b/www/src/components/repos/Repositories.js
@@ -185,7 +185,7 @@ export function RepositoryList({ repositores: { edges, pageInfo }, fetchMore, pu
     <Scroller
       id="repositories"
       edges={Array.from(chunk(edges, columns))}
-      style={{ overflow: 'auto', height: '100%', width: '100%'}}
+      style={{ overflow: 'auto', height: '100%', width: '100%' }}
       emptyState={emptyState}
       mapper={chunk => (
         <Box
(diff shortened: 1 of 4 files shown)


www/src/components/repos/Repositories.js
```diff
@@ -185,15 +185,15 @@ export function RepositoryList({ repositores: { edges, pageInfo }, fetchMore, pu
...
+      style={{ overflow: 'auto', height: '100%', width: '100%'}}
       emptyState={emptyState}
       mapper={chunk => (
         <Box
           key={chunk[0].node.id}
           direction="row"
           gap="small"
           fill="horizontal"
-          margin={{ bottom: 'small' }}
+          margin={{ bottom: 'small'}}
```
maciaszczykm, line 196: It seems to be not formatted.

michaeljguarino:
I think using a css import here is totally a valid approach actually

michaeljguarino (approved):

www/src/index.css
```diff
@@ -0,0 +1,14 @@
...
+
+* ::-webkit-scrollbar {
+    width: 6px;
+    background-color: #21242C;
+}
+
+* ::-webkit-scrollbar-thumb {
+    border-radius: 6px;
+    background-color: #2A2E37;
+}
```
dherault, line 14: You can use the Sylesheet key in the theme.tsx of the design-system to do this. It does the same, just to answer your question.

PR #321, opened 2022-07-06
Fix dropdown end icon and Lint

dherault:
## Summary

image
Resolves  ENG-245

## Checklist


www/src/components/repository/InstallDropdownButton.js
```diff
@@ -206,6 +203,9 @@ function InstallDropdownButton({ recipes, ...props }) {
             setRecipe(event.target.value)
             setTab(0)
           }}
+          endIcon={(
```
floreks, line 206: Aren't we losing the original animation and arrow up icon while the dropdown is open?

dherault: Yes, but that can be fixed later in honorable

michaeljguarino (approved):

PR #335, opened 2022-07-07
Bump design system version and update package.json

maciaszczykm:
## Summary
Related to ENG-270.

- Updates filter button styling in marketplace
- Bumps design system version
- Updates package.json config including switch to yarn and moving config to dedicated files

## Test Plan
Tested manually.

## Checklist


www/package.json
```diff
@@ -3,32 +3,15 @@
...
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
+  "engines": {
+    "node": ">=16.15.0",
+    "yarn": ">=1.22.0"
```
DavidSpek, line 14: I thought the latest version of Yarn is 3.2.0.

DavidSpek: Correction, 3.2.1. See https://yarnpkg.com/getting-started/install

DavidSpek: I'm no frontend guy so whatever makes sense haha

maciaszczykm: Done.

www/package.json
```diff
@@ -121,8 +104,5 @@
     "eslint": "8.19.0",
     "eslint-config-pluralsh": "3.1.0",
     "postcss-import": "8.2.0"
-  },
-  "peerDependencies": {
-    "react-transition-group": ">=4.4.2"
```
maciaszczykm, line 126: @dogmar it looks that you've added it but with the changes made here I've decided to put it back with all dependencies. It looks like we are using it directly in the code. Please let me know if you would rather not change it.

michaeljguarino (commented):
Is the `.yarn` folder supposed to be in this diff?

michaeljguarino (commented):
We'll also need to modify the dockerfile if we're goign to be using yarn again

maciaszczykm:
>Is the .yarn folder supposed to be in this diff?

Yes, see pt 6: https://yarnpkg.com/getting-started/migration#step-by-step

maciaszczykm:
@michaeljguarino I've restored previous version of Dockerfile that supports yarn. Docker build passed locally. PTAL.

DavidSpek:
> Is the .yarn folder supposed to be in this diff?

@michaeljguarino Yes, there are a couple files in there that you want to keep in the git history. I ran into this with the kubricks frontend as well.

floreks (approved):

DavidSpek:
@maciaszczykm It looks like the yarn version in the Dockerfile is not berry.

```console
Step 6/13 : RUN yarn
 ---> Running in 566bcb50e9dc
yarn install v1.22.19
```

PR #341, opened 2022-07-08
Generate typescript types for our GraphQL API

DavidSpek:
## Summary
Since our [Design System](https://github.com/pluralsh/design-system) is in TypeScript, I thought having the TS types for our GraphQL API might be handy at some point. I'm not sure if this should be saved here or in a separate package, but this is more meant as a reference for when we want to have the TS types.


floreks:
https://github.com/pluralsh/plural/pull/352 should actually allow us to start using typescript in our codebase and benefit from those types.

DavidSpek:
@floreks If this is something we want to start using in the current form I just rebased and updated the types.

www/package.json
```diff
@@ -12,7 +12,8 @@
     "fix": "yarn fix:js && yarn fix:css",
     "fix:js": "eslint ./src --fix",
     "fix:css": "stylelint src/**/*.css --fix",
-    "unimported": "npx unimported"
+    "unimported": "npx unimported",
+    "graphql-codegen": "graphql-codegen --config codegen.yml"
```
floreks, line 16: For consistency of naming use `graphql:codegen`.

maciaszczykm (approved):

maciaszczykm (approved):

PR #352, opened 2022-07-11
ENG-300 - Configure unit tests

floreks:
## Summary

This is a bit bigger change then initially expected:
- configured jest and added initial test
- bumped `react` and `react-dom` to `v18` and related packages to latest versions for compatibility
- configured github action to run unit tests
- removed lodash dependencies for single packages and used per package imports instead
- added initial `tsconfig.json`

We can now start using `tsx` extension for any source files in the codebase. If required `@types/xxx` packages can be installed to further improve type checking by the IDE. I'd start slowly rewriting components 1 by 1. 

**NOTE:** This should be carefully tested as react bump is a major change. I have tried to check everything but it would be good to get others to test too.

### Caveats

There are couple of things worth mentioning.

#### `resolutions` in the `package.json`
This is unfortunately required to force other dependencies to use the same react core version. Without that there are multiple versions used and react basically breaks. Ref: https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react

#### `.eslintrc.yml`
I am working on a PR to extract those changes to our common eslint config package.

## Checklist


.github/workflows/www.yaml
```diff
@@ -14,6 +14,28 @@ on:
       - ".github/workflows/www.yaml"
       - "www/**"
 jobs:
+  unit:
+    name: "Unit Tests"
```
maciaszczykm, line 18: If we are using name here, can you add one for lint too?

michaeljguarino (approved):

.github/workflows/www.yaml
floreks: Done

dherault (approved):

maciaszczykm (approved):

PR #362, opened 2022-07-14
Adjust fonts across the app

maciaszczykm:
## Summary
Fixes ENG-325.

## Test Plan
Tested manually.

## Checklist

### DIFF
--- a/www/src/theme.js
+++ b/www/src/theme.js
@@ -257,7 +257,7 @@ export const DEFAULT_THEME = {
       },
     },
     font: {
-      family: 'Monument',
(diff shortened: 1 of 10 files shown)


www/src/theme.js
```diff
@@ -257,7 +257,7 @@ export const DEFAULT_THEME = {
       },
     },
     font: {
-      family: 'Monument',
+      family: 'Inter',
```
maciaszczykm, line 260: It looks like in most of the new designs Inter font is being used. Monument is still used in buttons etc. but this comes from the design system components.

PR #375, opened 2022-07-19
Add "Edit" page to Repository tab

dogmar:
https://linear.app/pluralsh/issue/ENG-337/add-edit-page-to-repository-tab


www/src/components/repository/RepositoryEdit.tsx
```diff
@@ -0,0 +1,677 @@
...
+    update: updateFormState,
+    // reset: resetFormState,
+  } = useUpdateState<FormState>(
+    useMemo(
+      () => ({
+        name,
+        description,
+        category,
+        oauthUrl:
+          typeof oauthSettings?.uriFormat === 'string'
```
michaeljguarino, line 188: couldn't this just be `oauthSettings?.uriFormat || ''`?  Or is the typeof clause doing some work to hint ts to it being a string

www/src/components/repository/RepositoryEdit.tsx
```diff
@@ -0,0 +1,677 @@
...
+              authMethod: formState.oauthMethod,
+            }
+            : null,
+        ...(iconUpdate.file ? { icon: iconUpdate.file } : {}),
+        tags: formState.tags.map(t => ({ tag: t.name })),
+        private: formState.private,
+      },
+    },
+    update: (cache, { data: { updateRepository } }) => {
+      const prev: any = cache.readQuery({
```
michaeljguarino, line 251: is this cache update necessary, my guess is apollo will just handle it?  If it is, we have a helper `updateCache` that could compress some of the logic as well

www/src/components/repository/RepositoryEdit.tsx
```diff
@@ -0,0 +1,677 @@
...
+    file: File | null;
+    previewUrl: string | null;
+  }>({ file: null, previewUrl: icon || null })
+
+  const [, setSuccess] = useState(false)
+  const [, setError] = useState<ApolloError>()
+
+  const tagSearchRef = useRef<any>(null)
+
+  const [mutation, { loading }] = useMutation(UPDATE_REPOSITORY_MUTATION, {
```
michaeljguarino, line 231: you can actually just get the error by doing `const [mutation, { loading, error }]` here

www/src/components/repository/queries.js
```diff
@@ -18,6 +18,18 @@ export const CREATE_REPOSITORY_MUTATION = gql`
   ${RepoFragment}
 `
 
+export const UPDATE_REPOSITORY_MUTATION = gql`
+  mutation UpdateRepository($repositoryId: ID!, $attributes: RepositoryAttributes!) {
+    updateRepository(repositoryId: $repositoryId, attributes: $attributes) {
+      ...RepoFragment
+      tags {
+        name: tag
```
michaeljguarino, line 26: It might be best to not do this rename so the cache is consistent elsewhere that doesn't do that

www/src/components/repository/RepositoryEdit.tsx
```diff
@@ -0,0 +1,677 @@
...
+        category: formState.category,
+        oauthSettings:
+          formState.oauthUrl && formState.oauthMethod
+            ? {
+              uriFormat: formState.oauthUrl,
+              authMethod: formState.oauthMethod,
+            }
+            : null,
+        ...(iconUpdate.file ? { icon: iconUpdate.file } : {}),
+        tags: formState.tags.map(t => ({ tag: t.name })),
```
michaeljguarino, line 246: not doing this tag -> name switch might also eliminate the need for this map call

www/src/components/repository/RepositoryEdit.tsx
maciaszczykm: We can also use `${oauthSettings?.uriFormat || ''}` to make sure that it is a string.

maciaszczykm: To make sure that it is a string we can use:

```
`${oauthSettings?.uriFormat || ''}`
```

www/src/components/repository/RepositoryEdit.tsx
```diff
@@ -0,0 +1,677 @@
...
+  }, [initialState])
+
+  const hasUpdates = useMemo(() => {
+    for (const [prop, value] of Object.entries(state)) {
+      const curVal = !value && typeof value === 'string' ? null : value
+      const initialVal =
+        !initialState[prop] && typeof value === 'string'
+          ? null
+          : initialState[prop]
+      if (!isEqual(curVal, initialVal)) {
```
floreks, line 88: Wouldn't using lodash `isEqual` in general work here? It can do a deep object comparison.

www/src/components/repository/RepositoryEdit.tsx
```diff
@@ -0,0 +1,677 @@
...
+    initialState: { ...initialState },
+  }
+}
+
+function RepoIcon({
+  src = null,
+  onClick,
+}: {
+  src: string | null;
+  onClick: EventHandler<any>;
```
floreks, line 110: Any reason not to use `unknown` instead of `any`?

www/src/components/repository/RepositoryEdit.tsx
dogmar: I had a bunch of nulls in initialState that I needed to match empty strings, but I fixed by just making sure initialState started with empty strings.

www/src/components/repository/RepositoryEdit.tsx
dogmar: This switch was happening all over the app, but I changed it on those places as well.

www/src/components/repository/RepositoryEdit.tsx
floreks: That's even better. Great :slightly_smiling_face:

floreks (approved):

PR #376, opened 2022-07-20
Redesign packages screen

maciaszczykm:
## Summary
Closes ENG-316.

This pull request updates all views for charts and terraform charts. I've already done most of the layer changes, updated part of component styling, removed duplicated code but still quite a lot is missing.

## Test Plan
Tested manually.

## Checklist


www/src/components/repos/packages/ChartReadme.js
```diff
@@ -0,0 +1,15 @@
...
+
+import ChartContext from '../../../contexts/ChartContext'
+
+import { MARKDOWN_STYLING } from '../Chart'
+
+export default function ChartReadme() {
+  const { current } = useContext(ChartContext)
+
+  return (
+    <Box pad="large"><Markdown components={MARKDOWN_STYLING}>{current?.readme || 'n/a'}</Markdown></Box>
```
michaeljguarino, line 13: we can use react routers `Outlet` component to manage this a bit smoother like in other views

maciaszczykm: [I've used it already in a way like it is done for repositories](https://github.com/pluralsh/plural/pull/376/files#diff-8ca897a150f2157542874fde88bd45e021a3f748f1f1a3d38ba6b0fdf88c761bR291). I'm not sure if that's enough or you would like to move the `Outlet` placement or add another one here for different purpose. Can you explain?

michaeljguarino: Oh I see you are structuring it right, where is the Outlet component injected in the component tree?  I just couldn't find it in the diff anywhere

michaeljguarino:
We should also have this pr be based on the `packages-redesign` branch

www/src/components/repos/packages/ChartReadme.js
maciaszczykm: It's in Plural.js if i understand you correctly. There are only two of them at the moment as it's still in progress.

www/src/components/repository/RepositoryDescriptionMarkdown.js
```diff
@@ -35,7 +35,7 @@ function getLastStringChild(children) {
   return lastChild
 }
 
-function MdPre({ children, ...props }) {
+export function MdPre({ children, ...props }) {
```
michaeljguarino, line 38: could we move this to a place where it's a bit clearer that it's reusable (eg the utils directory?)

maciaszczykm: I've switched to another component already and will remove that export tomorrow.

maciaszczykm:
You can review it already, once it will get approved I will merge and continue to work from another branch. I've created subissues for better progress tracking: https://linear.app/pluralsh/issue/ENG-316/redesign-packages-screen

Some screens from current state:
Zrzut ekranu 2022-07-22 o 12 18 58
Zrzut ekranu 2022-07-22 o 12 19 14
Zrzut ekranu 2022-07-22 o 12 19 20
Zrzut ekranu 2022-07-22 o 12 19 45

www/src/components/repos/queries.js
```diff
@@ -195,7 +195,7 @@ export const CHART_Q = gql`
       }
       installation { ...ChartInstallationFragment }
     }
-    versions(chartId: $chartId, first: 10, after: $cursor) {
+    versions(chartId: $chartId, first: 15, after: $cursor) {
```
maciaszczykm, line 198: I'll switch back to 10 once the button to load more will be done. 100 cannot be used because it scroll doesn't show inside select panel.

michaeljguarino (approved):
Cool, good progress, let's get this merged and continue from there

PR #377, opened 2022-07-20
Properly resize cloud shell

michaeljguarino:
## Summary

Use the react-resize-detector module to send resize events and plumb the fitAddons dimensions in response.  We add a small debounce to prevent this from spamming too heavily.

## Test Plan
verified manually in cloud shell

## Checklist


www/src/components/shell/Terminal.js
```diff
@@ -67,14 +69,19 @@ export function Shell({ shell }) {
...
     channel.push('resize', { width: cols, height: rows })
-  }
+  }, [channel])
 
-  function handleData(text) {
-    channel.push('command', { cmd: text })
-  }
+  const { ref } = useResizeDetector({ onResize: debounce(() => {
+    if (!channel) return
+    console.log('resizing')
```
maciaszczykm, line 79: Is it a debug leftover?

www/package.json
```diff
@@ -75,6 +75,7 @@
     "react-icons": "4.4.0",
     "react-json-view": "1.21.3",
     "react-portal": "4.2.2",
+    "react-resize-detector": "^7.1.2",
```
maciaszczykm, line 78: I think it will be safer to stick to exact versions, without `~` or `^`.

www/src/components/shell/Terminal.js
michaeljguarino: yeah can rm

www/package.json
michaeljguarino: sure, will change

maciaszczykm: You need to run yarn again as lock file still has old version inside.

maciaszczykm (approved):

floreks (approved):

PR #378, opened 2022-07-20
Grid layout

dherault:
## Summary

image

Resolves ENG-336

Implements the new grid layout on the repository page. Creates the groundwork for the marketplace implementation.

## Checklist


dherault:
image

www/src/components/repository/RepositorySideCar.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <InstallationConfiguration
+        open={open}
+        setOpen={setOpen}
+        installation={installation}
+      />
+    </>
+  )
+}
+
+function RepositorySideCar({ ...props }) {
```
michaeljguarino, line 52: Why are you destructuring props like this?

www/src/components/repository/Repository.js
```diff
@@ -34,8 +32,8 @@ function Repository() {
       <Flex
         // These mp values are to align the looping logo with the previous looping logo.
         // Reload the page on /repository/foo to see it in action.
-        pt={24.25}
-        ml={-5}
+        paddingTop={24.25 * 16}
```
michaeljguarino, line 35: * 16?

www/src/components/repository/RepositorySideCar.tsx
dherault: OH that's because it had some other props during dev. Let me fix that.

www/src/components/repository/Repository.js
dherault: Yes, pt={24.25} means 24.25*16 padding.

michaeljguarino: ah i see

michaeljguarino:
This is going to be a layout that is used for all the packages screens as well as potentially a few others (marketplace?).  it would be nice if there were a simple orthogonal implementation of it, that's used here, but also the other places so it's easier to maintain/update in the future.

dherault:
Actually the screens are so different that it is almost impossible, I thought it through. If you find a better alternative please let me know.

dherault:
The margins are different based on the screens. That's half of the issue, the other half has to do with breakpoints.

dogmar:
I thought the grid was designed to be consistent across almost all pages. Maybe not for marketplace, but I know it should work for the new Accounts pages as well.

dherault:
Let me figure something out

floreks:
What do you think about adding some kind of layout components to the design system that would handle the grid layout, width and breakpoints, and we would then reuse them here and in console.

I am not sure if there is a global way to handle everything since every view is further split into 3 sections (sidenav, content, sidecar) and it has to be applied on a per view basis anyway.

dherault:
Done.

dherault:
For the lint errors the linter is not yet compatible with the typescript config, that my next PR

michaeljguarino (approved):

.vscode/settings.json
```diff
@@ -2,8 +2,10 @@
   "editor.codeActionsOnSave": {
```
floreks, line 2: Could we actually add the IDE-related files to `.gitignore` and remove them from the repo?

www/src/components/layout/ResponsiveLayout.tsx
```diff
@@ -0,0 +1,46 @@
+import { Div, Flex } from 'honorable'
```
floreks, line 1: We should probably move it eventually to the design system.

PR #384, opened 2022-07-25
Add typescript linter

dherault:
## Summary

Add new linter config and lint the app.

## Checklist


dogmar:
There needs to be some space between the content and the scrollbar on the Readme page. I'd also say either make sure the scrollbar doesn't extend into the page title area or don't have the page title be sticky.

Screen Shot 2022-07-25 at 3 10 20 PM

michaeljguarino (approved):
The changes are non-breaking right?

www/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  extends: [
+    '@pluralsh/eslint-config-typescript',
```
floreks, line 3: That's actually a bit weird that it works for JS|JSX files since the config only supports files with extensions `ts` and `tsx`.

www/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
...
+  ],
+  globals: {
+    JSX: true,
```
floreks, line 6: I guess we can add that directly to the global config instead.

www/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
...
+  },
+  rules: {
```
floreks, line 8: Is this needed?

www/package.json
```diff
@@ -29,6 +29,7 @@
     "@nivo/geo": "0.79.1",
     "@nivo/line": "0.79.1",
     "@octokit/core": "3.6.0",
+    "@pluralsh/eslint-config-typescript": "^2.2.0",
```
floreks, line 32: Remove `^` to use the strict version.

www/src/components/Breadcrumbs.js
```diff
@@ -1,4 +1,6 @@
-import { createContext, useContext, useMemo, useState } from 'react'
+import {
+  createContext, useContext, useMemo, useState,
+} from 'react'
```
maciaszczykm, line 3: @floreks Does `@pluralsh/eslint-config-typescript` set max line width? It looks like lines are broken quite fast and in my opinion this makes code a bit harder to read. Can we change that?

@pluralsh/frontend What do you think?

maciaszczykm: This is not related to the PR itself. Pinged you just for discussion.

floreks: There are basically no custom rules since it only extends the airbnb configs. This might also be another rule that checks a number of variables/method parameters. I think the default is 3 and that's why it breaks at 4.

www/package.json
dherault: Done

www/.eslintrc.js
dherault: Done

www/.eslintrc.js
dherault: Sure!

PR #387, opened 2022-07-26
ENG-310 - Redesign cluster view

floreks:
## Summary

[Linear](https://linear.app/pluralsh/issue/ENG-310/cluster-refinement)
[Figma](https://www.figma.com/file/DnikcwIxHK5HDXqbriI0Yf/Clusters-Tab?node-id=68%3A7884)

Mostly done, only waiting for the select component from the design system. I haven't removed the `clusters.js` file as it might still be useful until this is fully finished. We should be able to merge this to the branch and add select when it's ready.

Zrzut ekranu z 2022-07-26 15-50-41

## Test Plan

Locally

## Checklist


www/src/components/clusters/ClustersSidecar.tsx
```diff
@@ -0,0 +1,61 @@
...
+          },
+        }}
+      >
+        Console
+      </Button>
+      <Sidecar heading="Metadata">
+        <SidecarItem heading="Cluster name">
+          {queue.name}
+        </SidecarItem>
+        <SidecarItem heading="Git url">
```
maciaszczykm, line 41: Should we switch to `URL`?

www/src/contexts/QueueContext.ts
```diff
@@ -0,0 +1,3 @@
+import { createContext } from 'react'
+
+export default createContext < any >({})
```
maciaszczykm, line 3: Can we provide interface with types for the context?

www/src/components/clusters/ClustersSidecar.tsx
floreks: Not sure. I have followed the design here.

www/src/contexts/QueueContext.ts
floreks: We would need https://github.com/pluralsh/plural/pull/341 to get merged first since those are the API types. I have manually added some partial types just to simplify the development, but I'd prefer to use generated types here.

maciaszczykm: Ok, sure.

PR #393, opened 2022-07-27
My profile refinement

maciaszczykm:
## Summary
Related to [ENG-307](https://linear.app/pluralsh/issue/ENG-307/my-profile-refinement).
Closes ENG-426.

[Figma design](https://www.figma.com/file/Y8iNQX9MTG2uZIjNelBroF/My-Profile?node-id=245%3A14643).

Used packages-redesign as a base branch here but I can switch to some other branch if you would like to.

Zrzut ekranu 2022-07-29 o 14 06 39
Zrzut ekranu 2022-07-29 o 14 06 45
Zrzut ekranu 2022-07-29 o 14 07 09

## Test Plan
Tested manually.

## Checklist


www/src/components/utils/Header.js
```diff
@@ -0,0 +1,26 @@
+import { Box } from 'grommet'
+import { Flex, H1, Span } from 'honorable'
+
+export function Header({ header, description, children }) {
```
floreks, line 4: Is it used? I didn't find any usage in the new code.

maciaszczykm: It's imported in `plural/www/src/components/account/AccountAttributes.js` and `plural/www/src/components/repository/OIDCProvider.js`.

dogmar (changes requested):
Need to add [profile picture switch/delete functionality](https://www.figma.com/file/Y8iNQX9MTG2uZIjNelBroF/My-Profile?node-id=247%3A14484). 

Bottom of profile picture doesn't align with Upload button. I think if you switch out the Avatar component for the new IconFrame from the design system, that will fix it. 
Screen Shot 2022-07-29 at 8 59 47 AM

dogmar:
On the security screen, if you enter a valid password, then select all of the text in the password field and delete it at once, the hint still says "Valid password" even though the field is empty.

maciaszczykm:
> Bottom of profile picture doesn't align with Upload button. I think if you switch out the Avatar component for the new IconFrame from the design system, that will fix it.

Nice catch! `Stack` was missing height property and that was causing the issue. I've tried using `IconFrame` but it has some issues with aspect ratios. When image is not a square, then it forces it and it looks worse than when using `Avatar`.

Zrzut ekranu 2022-08-1 o 11 48 01

Also, I think we cannot replace `Avatar` with `IconFrame` as logic for creating default picture is part of the first one.

maciaszczykm:
> On the security screen, if you enter a valid password, then select all of the text in the password field and delete it at once, the hint still says "Valid password" even though the field is empty.

Nice catch. I didn't change that part yet as I need some input from design team but anyways it doesn't work as expected because `validation` method is not called when doing `CTRL` + `A` and then `DELETE`. I've opened [a new bug to track this](https://linear.app/pluralsh/issue/ENG-439/validation-is-not-called-when-doing-ctrl-a-and-then-delete-in).

maciaszczykm:
> Need to add [profile picture switch/delete functionality](https://www.figma.com/file/Y8iNQX9MTG2uZIjNelBroF/My-Profile?node-id=247%3A14484).

I wanted to add this from separate PR/[subissue](https://linear.app/pluralsh/issue/ENG-426/add-a-way-to-deletereset-profile-picture-to-default) but there is no need to split as it's small change. Already done.

@dogmar Thanks for precise review, PTAL.

michaeljguarino (approved):

michaeljguarino:
@dogmar could you give this a second look, I believe your review is the only thing blocking

www/src/components/utils/Header.js
floreks: It's imported from `src/components/profile/Header.js` there, not this one from what I can see.

floreks:
We'd probably need to add a custom upload modal with crop functionality to be able to use icon frame or check if the picture has the proper ratio before upload.

www/src/components/utils/Header.js
maciaszczykm: https://github.com/pluralsh/plural/blob/9b66cc61305f31398bb324871bc68c5acdd72cd0/www/src/components/account/AccountAttributes.js#L10-L11

maciaszczykm: You're right, good catch. I've started this to move Header to utils as it's used not only in the profile but across whole app. Probably switched to something and forgot to finish.

dogmar (approved):
Thanks for addressing all my comments. Looks good!

PR #395, opened 2022-07-28
Clusters. Domain is a direct link to the console, if it exists

git-toni:
## Summary

After briefly trying the platform I found it would be quite helpful to have a direct link to the cluster console. Plus it's an small way to get to know the codebase :slightly_smiling_face: 
Done:

- Upgrade elixir/otp in `.tool-versions` (I can undo this if you prefer, granted it's a bit out of place)
- Make clusters' domains a new-tab link.

Before
image

After
image

## Labels

`enhancement` 

## Checklist


.tool-versions
```diff
@@ -1,2 +1 @@
-erlang 23.2
```
michaeljguarino, line 1: would you mind reverting this back for now?  Our local dev is still mostly elixir 1.11

michaeljguarino (commented):
Thanks for this contribution, just a few minor changes.  Also did you not notice the view console button, which I believe does something pretty similar or does this UX seem clearer to you?

.tool-versions
michaeljguarino: The elixir 1.12/OTP 24 migration is still pending for us, there were a few breaking changes in some dependencies that were a huge pain to figure out but we def want to do it.

www/src/components/clusters/Clusters.js
```diff
@@ -49,6 +49,21 @@ function DataEntry({ name, value, children }) {
 }
 
 function ClusterDescription({ q }) {
+  function DomainLink() {
+    if (q.domain == null || q.domain === '') return (<i>Empty</i>)
+
+    return (
+      <a
```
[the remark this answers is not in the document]
git-toni: Aah, my bad sorry :facepalm:  I went to check honorable's [storybook](https://storybook.honorable.design/) and couldn't find an anchor tag component. I see it's used around the codebase, thanks, fixed!

www/src/components/clusters/Clusters.js
```diff
@@ -49,6 +49,21 @@ function DataEntry({ name, value, children }) {
 }
 
 function ClusterDescription({ q }) {
+  function DomainLink() {
+    if (q.domain == null || q.domain === '') return (<i>Empty</i>)
```
michaeljguarino, line 53: this clause will technically never hit because of api validation but it doesn't hurt to be safe

michaeljguarino (approved):

www/src/components/clusters/Clusters.js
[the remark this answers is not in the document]
git-toni: UI Proof :slightly_smiling_face: 
image

.tool-versions
git-toni: Absolutely, it was a bit too much to add this into this unrelated PR :D

git-toni:
> Thanks for this contribution, just a few minor changes. Also did you not notice the view console button, which I believe does something pretty similar or does this UX seem clearer to you?

Oh I see it now, sorry. Feel free to close this PR :pray: I guess my brain is used to find an "inline" link with the classic ascii icon.

samweaver:
@git-toni thanks so much for the contribution, awesome to see!! 🥳

floreks:
@git-toni Nice job :) Just FYI we are working on a cluster view redesign. Check out PR #384 and let us know if you think it's more intuitive.

git-toni:
> Nice job :) Just FYI we are working on a cluster view redesign. Check out PR #387 and let us know if you think it's more intuitive.

Thank you @floreks, it's nothing! Just an excuse to play with the codebase 🙂

The re-design looks lit! Maybe I'd highlight the button a bit more, could be with a different background color; but that's a matter of taste. Alternatively making it the top sibling of `Git url` with the style of an anchor also would be quite intuitive for me.  :relaxed: 
image


If you are all in the middle of re-designing I can happily close this tiny PR here, so you can safe one release cycle etc for such a small change. Let me know!

samweaver:
This is good feedback from @git-toni that we should share with design team – @floreks perhaps you can chat with Nick about it and get his perspective?

floreks:
> This is good feedback from @git-toni that we should share with design team – @floreks perhaps you can chat with Nick about it and get his perspective?

Will do! Let's see if we can make it even better.

PR #399, opened 2022-07-29
ENG-311 - Repository refinement

floreks:
## Summary

Ref: [Linear](https://linear.app/pluralsh/issue/ENG-311/repository-refinement) | [Figma](https://www.figma.com/file/kZ69OGXTQe3Gu0wzGdDyoO/Marketplace-Tab?node-id=909%3A25834)

#### Marketplace
- Add verified badge support
- Use large repository cards for featured section
#### Repository view
- Updated views to use `PageTitle` from the design system
- Redesigned the `OpenID Connect` view
  - Use `ContentCard` as base for form data
  - Add `Client ID` and `Client secret` inputs and disable them as they should not be changed
  - Update other components to match the design
- Update Readme view and markdown styling according to the specs
- Redesigned the `Tests` view as per spec
- Fixed an issue with the infinite scroller where it was not loading more elements when the initial height of the table was bigger than the height of initially loaded elements (no visible scroll). It is not automatically trying to load more elements if the scrollbar is not visible to fill the table.
#### Toasts
- Add 5s close timeout to toasts
- Move toast to the right bottom corner
- Toast will not add an invisible overlay that blocks user actions now and clicking outside of toast will not trigger any action.

and more... :slightly_smiling_face:

## Test Plan

Locally

## Checklist


www/src/components/account/Typeaheads.js
```diff
@@ -140,26 +140,24 @@ const FETCHER = {
...
-  const { placeholder, label } = TEXT[type]
   const fetch = fetcher || FETCHER[type]
 
   return (
     <TagInput
       noborder
       placeholder={placeholder}
       hint={hint}
-      icon={ICONS[type]}
+      icon={type ? ICONS[type] : null}
```
maciaszczykm, line 154: Is it needed? Previously `undefined` was returned instead.

maciaszczykm (approved):

www/src/components/repository/RepositoryDescription.js
```diff
@@ -27,7 +27,6 @@ export default function RepositoryDescription() {
         flexGrow={1}
         overflowY="auto"
         paddingBottom="xlarge"
-        marginTop="medium"
```
maciaszczykm, line 25: Why it is removed? Was it moved somewhere else?

www/src/components/repository/RepositoryDeployments.js
```diff
@@ -111,7 +111,6 @@ function RepositoryDeployments() {
       />
       <Div
         fill
-        marginTop="medium"
```
maciaszczykm, line 107: As in https://github.com/pluralsh/plural/pull/399/commits/fd8af7eb8ff0b413e89abaa82cae3ef8f1271ede#r940113280.

www/src/components/repository/RepositoryDescription.js
floreks: That's because `PageTitle` component already adds 24px of bottom padding under horizontal line and adding an additional 16px would make it 32. According to the design, I have been following it should be 24px for all repository views.

Zrzut ekranu z 2022-08-08 13-07-12

Zrzut ekranu z 2022-08-08 13-10-23

maciaszczykm: Ok, good. This was missing few changes before. Looks to be fine now. Thanks!

maciaszczykm (approved):

PR #400, opened 2022-08-01
Create community info schema for repositories

maciaszczykm:
## Summary
Closes [ENG-332](https://linear.app/pluralsh/issue/ENG-332/create-community-info-schema-for-repositories).

## Checklist

### DIFF
--- a/apps/core/lib/core/schema/repository.ex
+++ b/apps/core/lib/core/schema/repository.ex
@@ -58,7 +58,7 @@ defmodule Core.Schema.Repository do
       field :discord, :string
       field :slack,   :string
       field :twitter, :string
-      field :videos,  :map
+      field :videos,  {:array, :string}
     end
 
     belongs_to :integration_resource_definition, ResourceDefinition, on_replace: :update
--- a/apps/core/lib/core/schema/repository.ex
+++ b/apps/core/lib/core/schema/repository.ex
@@ -54,6 +54,13 @@ defmodule Core.Schema.Repository do
       field :url,  :string
     end
 
+    embeds_one :community, Community, on_replace: :update do
+      field :discord, :string
+      field :slack,   :string
+      field :twitter, :string
+      field :videos,  :map
+    end
+
     belongs_to :integration_resource_definition, ResourceDefinition, on_replace: :update
     belongs_to :publisher, Publisher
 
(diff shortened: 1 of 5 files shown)


maciaszczykm:
@michaeljguarino Can you take first look? I'll try to add some tests now.

apps/core/lib/core/schema/repository.ex
```diff
@@ -54,6 +54,13 @@ defmodule Core.Schema.Repository do
       field :url,  :string
     end
 
+    embeds_one :community, Community, on_replace: :update do
+      field :discord, :string
+      field :slack,   :string
+      field :twitter, :string
+      field :videos,  {:array, :string}
```
michaeljguarino, line 61: one thing that might be a good idea here is to make this a structure with video type (since embedding youtube vs vimeo vs standard video urls is a bit different).  If there's a prebaked video component that can discover this from url alone, it's unnecessary though

maciaszczykm: https://github.com/CookPete/react-player seems to be perfect for this. WDYT?

maciaszczykm:
@michaeljguarino I've added tests, PTAL.

apps/core/test/services/repositories_test.exs
```diff
@@ -57,9 +70,18 @@ defmodule Core.Services.RepositoriesTest do
...
+        name: "piazza",
+        community: %{
+          twitter: "twitter.com/piazza",
+          videos: ["https://www.youtube.com/watch?v=dQw4w9WgXcQ"]
+        }
+      }, repo.id, user)
 
       assert updated.name == "piazza"
+      assert updated.community.twitter == "twitter.com/piazza"
+      assert length(updated.community.videos) == 1
```
michaeljguarino, line 83: you should be able to just do `asset updated.community.videos == ["https://www.youtube.com/watch?v=dQw4w9WgXcQ"]` I believe

maciaszczykm: True, makes no sense to split it into 2 lines. Fixed.

maciaszczykm:
@michaeljguarino PTAL.

apps/core/lib/core/schema/repository.ex
michaeljguarino: Yeah agreed

michaeljguarino (changes requested):

PR #402, opened 2022-08-03
Fix view crash on cloud shell boot

floreks:
## Summary

This fixes the frontend part of the issue. It is not crashing anymore but the boot sequence requires view reload to properly start the initial session. This looks like a backend issue. 

## Test Plan

Delete shell pod via the console and open cloud shell view again. No crashing or errors on the frontend side.

## Checklist

### DIFF
--- a/www/src/components/shell/Terminal.js
+++ b/www/src/components/shell/Terminal.js
@@ -39,7 +39,7 @@ export function Shell({ shell }) {
     if (!xterm?.current?.terminal) return
 
     const term = xterm.current.terminal
-    const chan = socket.channel('shells:me', {})
+    const chan = socket.channel('shells:me')
 
     fitAddon.fit()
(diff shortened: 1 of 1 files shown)


www/src/components/shell/Terminal.js
```diff
@@ -161,21 +161,22 @@ export function Shell({ shell }) {
 
 export function Terminal() {
   const { data } = useQuery(CLOUD_SHELL_QUERY, { pollInterval: 5000, fetchPolicy: 'cache-and-network' })
+  const { shell } = data
+  const { alive } = data?.shell || false
+  const { status } = data?.shell || undefined
```
maciaszczykm, line 166: `const { alive = false, status = undefined } = data?.shell` should do the same.

Do you need `undefined` here?

floreks: That's why I've had to split it
image

maciaszczykm (approved):

PR #409, opened 2022-08-09
Redesign package version picker

maciaszczykm:
## Summary
Fixes ENG-368.

[Design](https://www.figma.com/file/RGR6BUySI9EfxBXbLGDPMV/Packages?node-id=2%3A2).

Zrzut ekranu 2022-08-10 o 11 53 33 Zrzut ekranu 2022-08-10 o 11 51 39
Zrzut ekranu 2022-08-10 o 11 51 11

## Test Plan
Tested manually.

## Checklist


michaeljguarino:
For some reason I can't click any of the sidebar tabs in the preview link instance, not sure what the cause might be, perhaps some borked click handler

floreks (approved):

floreks:
I see the same issue when running locally. It seems that the invisible select popover element takes a lot of space and acts as an overlay that also blocks access to other UI elements.

maciaszczykm:
Yep, it looks like `display: none` is missing somewhere and invisible layer blocks access to elements underneath.

www/src/components/repos/common/misc.js
```diff
@@ -93,49 +96,84 @@ export function PackageHeader({ name, icon }) {
...
+                showExtra
+                chips={[
+                  v.id === installed?.version?.id && (
+                    <Chip
+                      severity="success"
+                      size="small"
+                    >
+                      Installed
+                    </Chip>
+                  ),
```
dogmar, line 150: This inserts `false` as the first element of the array if it's not the installed version. So the wrong number of chips gets displayed.

dogmar: This should work: `...(v.id === installed?.version?.id ? [(<Chip />)] : [])`

dogmar (changes requested):

www/src/components/repos/common/misc.js
maciaszczykm: Good catch, thanks.

floreks (approved):

www/src/components/repos/common/misc.js
```diff
@@ -93,49 +96,84 @@ export function PackageHeader({ name, icon }) {
...
+                showExtra
+                chips={[
+                  ...(v.id === installed?.version?.id ? (
+                    <Chip
+                      severity="success"
+                      size="small"
+                    >
+                      Installed
+                    </Chip>
+                  ) : []),
```
dogmar, line 150: Now this errors if the package is installed. Need to wrap the "Installed" `Chip` with `[]` for the spread operator to work.

dogmar (changes requested):

www/src/components/repos/common/misc.js
maciaszczykm: 🤦🏼 Fixed.

maciaszczykm:
@dogmar PTAL.

michaeljguarino (approved):
Looks great

PR #413, opened 2022-08-09
Fix some broken repo Readme images

dogmar:
If the repo’s main branch was anything other than `master`, URLs for readme images would be incorrect. This fixes that by pulling in the new `main_branch` field for repos.

### DIFF
--- a/www/src/components/repository/queries.js
+++ b/www/src/components/repository/queries.js
@@ -37,6 +37,7 @@ export const REPOSITORY_QUERY = gql`
         name: tag
       }
       readme
+      main_branch
       git_url
       homepage
       license {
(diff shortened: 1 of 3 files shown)


maciaszczykm (approved):

www/src/components/repository/queries.js
```diff
@@ -37,6 +37,7 @@ export const REPOSITORY_QUERY = gql`
         name: tag
       }
       readme
+      main_branch
```
michaeljguarino, line 40: we should probably camel case these actually to match the conventions elsewhere (the gql api supports both patterns)

michaeljguarino (approved):

PR #420, opened 2022-08-12
feat(frontend): run nginx as non-root user + fix prod build

DavidSpek:
## Summary
Runs the frontend container as non-root.

## Checklist

### DIFF
--- a/www/nginx.conf
+++ b/www/nginx.conf
@@ -1,48 +1,37 @@
 server {
-    listen       80;
-    server_name  localhost;
-    #charset koi8-r;
-    #access_log  /var/log/nginx/host.access.log  main;
+    # listen on port 80
+    listen 8080;
+    # save logs here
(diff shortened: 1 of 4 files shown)


www/nginx.conf
```diff
@@ -1,48 +1,37 @@
 server {
-    listen       80;
-    server_name  localhost;
-    #charset koi8-r;
-    #access_log  /var/log/nginx/host.access.log  main;
+    # listen on port 80
+    listen 8080;
```
floreks, line 3: Should it be `listen 80;`?

DavidSpek: 80 is a privileged port and requires root (same for 443). Since this uses the non-privileged NGINX it listens on port 8080.

floreks: So it is just the comment

DavidSpek: yep

PR #421, opened 2022-08-12
ENG-327: Plural release process

DavidSpek:
## Summary
Copy the new release process of the Console to Plural. The actions will need to be modified to accommodate the multi-app building done for Plural. Other than that, most of the workflows from the console should be mostly just be copy/paste work.

## Checklist


.dockerignore
```diff
@@ -3,7 +3,6 @@ deps/
 charts/
 .vscode/
 log/
-.git/
```
michaeljguarino, line 6: why remove this from dockerignore?

.github/workflows/push-to-plural.yaml
```diff
@@ -0,0 +1,30 @@
+name: CD / Plural
+
+on:
+  push:
+    branches:
+      - "master"
+    paths:
+      - "plural/**"
+jobs:
+  deploy:
```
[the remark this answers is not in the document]
DavidSpek: There isn't really much that can be tested here since all that will have changed is the image tags in the helm values.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,204 @@
...
+      - uses: erlef/setup-beam@v1
+        with:
+          elixir-version: '1.11.4' # Define the elixir version [required]
+          otp-version: '23.1.5' # Define the OTP version [required]
+      - uses: azure/setup-helm@v1
+        with:
+          version: latest
+      - name: install plural cli
+        run: |
+          wget -O plural.o 'https://app.plural.sh/artifacts/plural/plural?platform=linux&arch=amd64'
```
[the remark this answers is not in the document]
DavidSpek: What is the CLI even used for here?

.dockerignore
DavidSpek: We need the `.git` directory when building the container since Elixir will get the version from the Git tag.

DavidSpek:
related to the `cleanup tools download` commit for the dockerfile. I tested all CLIs if they executed by running a built container locally and here's the `ls` output:

```console
bash-5.1# ls /usr/local/bin/
ct_run     dialyzer   epmd       erl        erlc       escript    goon       helm       plural     rebar3     run_erl    terrascan  to_erl     trivy      typer
```

.github/workflows/publish.yaml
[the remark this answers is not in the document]
michaeljguarino: We use it to extract docker images from helm charts/validate they can be templated on upload

.github/workflows/push-to-plural.yaml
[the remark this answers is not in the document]
michaeljguarino: doesn't it execute on a push straight to master?  If so we should still validate unit tests before any of this hits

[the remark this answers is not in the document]
DavidSpek: Only if the path is `plural/*`, and all this does is push the Plural artifact so no actual code changes will be included.

renovate.json
```diff
@@ -26,8 +26,41 @@
...
+    {
+      "fileMatch": ["^Dockerfile$"],
+      "matchStrings": [
+        "datasource=(?<datasource>.*?) depName=(?<depName>.*?)( versioning=(?<versioning>.*?))?\\s(?:ENV|ARG) .*?_VERSION=(?<currentValue>.*)\\s"
+      ],
+      "versioningTemplate": "{{#if versioning}}{{{versioning}}}{{else}}semver{{/if}}"
     }
   ],
   "labels": ["dependencies"],
+  "regexManagers": [
```
floreks, line 56: This looks identical to the regex manager above.

.dockerignore
michaeljguarino, line 6: we really shouldn't be including the git subfolder in our images

michaeljguarino: (if we can get around tagging the version with just an env var or something similar being there at build, I think that's fine)

DavidSpek: the .git folder actually isn't very large and it never makes it into the final image.

michaeljguarino: Oh that is true it's only used in the builder I suppose.  Still feel like the approach is going to bite us at some point (and github actions has a canonical env var for tag I believe)

DavidSpek: It hasn't caused problems with the console. An env var seems less stable than the git tags, also between environments.

apps/api/mix.exs
```diff
@@ -1,10 +1,45 @@
 defmodule Api.MixProject do
   use Mix.Project
 
+  defp version do
+    case :file.consult('hex_metadata.config') do
```
michaeljguarino, line 5: If we're going to go this route, we should at least ditch this clause, since there should never be a hex_metadata.config file in our setup (we never publish to hex)

apps/api/mix.exs
```diff
@@ -1,10 +1,45 @@
...
+      {:ok, data} ->
+        {"version", version} = List.keyfind(data, "version", 0)
+        version
+      _ ->
+        version =
```
michaeljguarino, line 10: we should be able to just use this clause

PR #426, opened 2022-08-16
Ensure email confirmation sent on email updates

maciaszczykm:
## Summary
Closes ENG-494.

## Test Plan
Added tests.

## Checklist


michaeljguarino (commented):
This looks good, but we should add some unit tests too

maciaszczykm:
@michaeljguarino Sure, I am on it as it's still in progress.

maciaszczykm:
@michaeljguarino PTAL. I've added tests and confirmed that it creates following reset tokens:

```
[email: "user-2@example.com", type: :email]
```
I'm not familiar with email setup yet, so please check if that is enough.

By the way, should we reset `email_confirmed` fields to be able to use them in the UI?

apps/core/lib/core/pubsub/protocols/fanout.ex
```diff
@@ -45,6 +45,15 @@ defimpl Core.PubSub.Fanout, for: Core.PubSub.UserCreated do
   end
 end
 
+defimpl Core.PubSub.Fanout, for: Core.PubSub.UserUpdated do
+  alias Core.Services.Users
+
+  def fanout(%{item: %{email_changed: true} = user}) do
+    Users.create_reset_token(%{type: :email, email: user.email})
+  end
+  def fanout(_), do: :ignore
```
michaeljguarino, line 54: this is a mega-nit, but we usually just return `:ok` for these

michaeljguarino (approved):

apps/core/lib/core/pubsub/protocols/fanout.ex
maciaszczykm: Ok, done.

michaeljguarino (approved):
feel free to merge whenever you're comfortable

maciaszczykm:
@michaeljguarino I've added reset for `email_confirmed` and `email_confirm_by`. PTAL, you can merge if it looks good.

apps/core/test/services/users_test.exs
```diff
@@ -66,6 +66,19 @@ defmodule Core.Services.UsersTest do
       assert_receive {:event, %PubSub.UserUpdated{item: ^updated}}
     end
 
+    test "email change is properly detected" do
+      user = insert(:user)
+
+      {:ok, updated} = Users.update_user(%{name: "changed"}, user)
+      assert updated.email_changed == false
```
michaeljguarino, line 73: you can technically just use `assert` and `refute` here although this obviously still works

michaeljguarino (approved):

apps/core/lib/core/pubsub/protocols/fanout.ex
```diff
@@ -45,6 +45,18 @@ defimpl Core.PubSub.Fanout, for: Core.PubSub.UserCreated do
   end
 end
 
+defimpl Core.PubSub.Fanout, for: Core.PubSub.UserUpdated do
+  alias Core.Services.Users
+
+  def fanout(%{item: %{email_changed: true} = user}) do
+    confirm_by = Timex.now() |> Timex.shift(days: 7)
+    Users.update_user(%{email_confirmed: false, email_confirm_by: confirm_by}, user)
```
michaeljguarino, line 53: this logic should be added to the changeset in `Core.Schema.User` actually

apps/core/test/services/users_test.exs
maciaszczykm: It's changed as it looks much cleaner that way.

apps/core/lib/core/pubsub/protocols/fanout.ex
maciaszczykm: It's done.

apps/core/lib/core/schema/user.ex
```diff
@@ -161,8 +161,13 @@ defmodule Core.Schema.User do
   end
 
   def set_email_changed(cs, model) do
+    confirm_by = Timex.now() |> Timex.shift(days: 7)
+
     case get_change(cs, :email) do
-      email when (not is_nil(email)) and (email != model.email) -> put_change(cs, :email_changed, true)
+      email when (not is_nil(email)) and (email != model.email) ->
```
michaeljguarino, line 167: oh i didn't notice this but the email != email will be covered by get_change (it shouldn't return a value if there's no change I believe).

Also it's usually better to match `email when is_binary(email)` since a positive match is more robust

maciaszczykm: Okay, I wasn't sure about it and that's why it's here. I'll update.

maciaszczykm:
@michaeljguarino PTAL.

apps/core/lib/core/schema/user.ex
```diff
@@ -159,6 +160,19 @@ defmodule Core.Schema.User do
...
     |> change_markers(password_hash: :password_change)
     |> cast_attachments(attrs, [:avatar], allow_urls: true)
+    |> set_email_changed(model)
+  end
+
+  def set_email_changed(cs, model) do
+    confirm_by = Timex.now() |> Timex.shift(days: 7)
+
+    case get_change(cs, :email) do
+      email when is_binary(email) and email != model.email ->
```
michaeljguarino, line 170: are you sure you need to do the `email != model.email` check?

maciaszczykm: @michaeljguarino You are right, it's redundant as `get_change` will return `nil` if email has not changed. Then it will not pass `is_binary` check. Fixed, PTAL.

michaeljguarino (approved):

PR #431, opened 2022-08-17
ENG-381 - New update experience and service-worker logic

floreks:
### Changes
- Major refactoring of our service worker logic
- Added `yarn start:prod` script to build and run prod version of our frontend locally as both service worker and new update experience logic will be disabled for dev.
- Updated `clusters` view to be able to benefit from apollo cache. It appears that we have to check for `data` instead of `loading`
to make it work.
- Removed update notification icon from the sidebar
- Added new info toast to inform about new update
- Updated logic that controls when to show information about available update
- Used `sessionStorage` instead of `localStorage` to store `gitCommit` information 
- Increased toast timeout to 10s

### Service worker
It turns out that the current service worker logic does not actually cache anything for us. It can be easily checked in the dev tools `Application -> Service Worker -> Network requests`. On `app.plural.sh` nothing is really taken from the service worker directly and the `Application -> Storage` tab shows the same. We've been relying only on built-in browser caching that can use both mem and disk cache to speed up the page load. I have added basic caching for the images and fonts that are part of our build.

As for the `sessionStorage` usage, I have decided to go with it since it is not shared between tabs. This way if the user has multiple tabs open, all of them will have their own "update" context and notifications will not be dismissed if the app gets updated only in a single tab.

#### app.plural.sh
image
image

#### Now
image
image

### New update experience
In order to avoid showing the toast notification when a user comes back to the application after some time (opening a new tab), I have added small workaround logic that will run only once when the application is loaded. A small caveat is that it is still somewhat visible to the user since it runs after the required data is loaded from the API. After that, it will periodically check if there is a new version available and if so, show the notification toast.

image

## Test Plan

Locally using prod build and serve.

## Checklist


www/src/components/layout/WithApplicationUpdate.js
```diff
@@ -1,48 +1,55 @@
...
-      setOpen(false)
-    }
-  }, [config, setOpen])
+      window.location.reload()
+    })
+  }, [config])
+  const reloadOnStale = () => (stale ? reloadApplication() : null)
+
+  // force rerender every 60s to check if version hasn't changed
+  useEffect(() => {
```
floreks, line 34: I am not 100% sure if this is needed. The config that is part of the plural context should trigger rerender here. I did use it to trigger the update logic when `sessionStorage` entry was changed by me manually. It was easier than mocking the graphql API response to trigger it. In the end, it shouldn't have any real impact on performance as it is quite small and triggered every 60s.

www/src/components/layout/WithApplicationUpdate.js
```diff
@@ -1,48 +1,55 @@
...
-  const stale = commit !== config.gitCommit
+    return () => clearInterval(interval)
+  }, [time, setStale, commit, config])
 
-  if (!stale || !open) return children({ reloadApplication, shouldReloadApplication: false })
+  // Run only once on first app load
+  // It enforces full app reload if app has been updated
+  useEffect(() => {
+    reloadOnStale()
+  }, [])
```
dherault, line 46: React Hook useEffect has a missing dependency: 'reloadOnStale'. Either include it or remove the dependency array.
GitHub is right here.

dherault (approved):

www/src/components/layout/WithApplicationUpdate.js
floreks: Including a dependency will trigger this effect on every dependency change and this is exactly what I wanted to avoid here. Right now it is only executed once on application load.

www/src/components/layout/WithApplicationUpdate.js
```diff
@@ -1,48 +1,55 @@
...
 
 import { PluralConfigurationContext } from '../login/CurrentUser'
 
 const COMMIT_KEY = 'git-commit'
 
 export const DEFAULT_COMMIT = 'plural-default-commit'
 
-export const getCommit = () => localStorage.getItem(COMMIT_KEY) || DEFAULT_COMMIT
-export const setCommit = sha => localStorage.setItem(COMMIT_KEY, sha)
+export const getCommit = () => sessionStorage.getItem(COMMIT_KEY) || DEFAULT_COMMIT
```
michaeljguarino, line 13: this could be done in a separate pr but we should actually inject the commit using an env var while building the docker container instead of storing it in local storage.  That will be a lot more reliable overall.

PR #453, opened 2022-08-23
ENG-534 - Set proper scrollable areas

dogmar:
Fixes issues with scrollable areas inside TabPanels. Also added some more remaining TabLists.


maciaszczykm (approved):
Looks good.

PR #454, opened 2022-08-24
ENG-381 - New update experience and service-worker logic

floreks:
### Changes
- Major refactoring of our service worker logic
- Added `yarn start:prod` script to build and run prod version of our frontend locally as both service worker and new update experience logic will be disabled for dev.
- Updated `clusters` view to be able to benefit from apollo cache. It appears that we have to check for `data` instead of `loading`
to make it work.
- Removed update notification icon from the sidebar
- Added new info toast to inform about new update
- Updated logic that controls when to show information about available update
- Used `sessionStorage` instead of `localStorage` to store `gitCommit` information 
- Increased toast timeout to 10s

### Service worker
It turns out that the current service worker logic does not actually cache anything for us. It can be easily checked in the dev tools `Application -> Service Worker -> Network requests`. On `app.plural.sh` nothing is really taken from the service worker directly and the `Application -> Storage` tab shows the same. We've been relying only on built-in browser caching that can use both mem and disk cache to speed up the page load. I have added basic caching for the images and fonts that are part of our build.

As for the `sessionStorage` usage, I have decided to go with it since it is not shared between tabs. This way if the user has multiple tabs open, all of them will have their own "update" context and notifications will not be dismissed if the app gets updated only in a single tab.

#### app.plural.sh
image
image

#### Now
image
image

### New update experience
In order to avoid showing the toast notification when a user comes back to the application after some time (opening a new tab), I have added small workaround logic that will run only once when the application is loaded. A small caveat is that it is still somewhat visible to the user since it runs after the required data is loaded from the API. After that, it will periodically check if there is a new version available and if so, show the notification toast.

image

## Test Plan

Locally using prod build and serve.

## Checklist


floreks:
Reopened as a new PR as the original PR https://github.com/pluralsh/plural/pull/431 was broken.

.github/workflows/firebase-hosting-pull-request.yml
```diff
@@ -2,14 +2,14 @@
 # https://github.com/firebase/firebase-tools
 
 name: Deploy to Firebase Hosting on PR
-'on': pull_request
+on: pull_request
```
dherault, line 5: I've always see that `on` with quotes, I'm not sure we should remove it. Does it work?

.github/workflows/firebase-hosting-pull-request.yml
```diff
@@ -2,14 +2,14 @@
...
 jobs:
   build_and_preview:
     if: '${{ github.event.pull_request.head.repo.full_name == github.repository }}'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: cd www && yarn install --frozen-lockfile && npm run build
+      - run: cd www && yarn install --immutable && CI=false yarn build
```
dherault, line 12: Why edit the CI?

www/src/components/utils/Toasts.js
```diff
@@ -63,3 +63,13 @@ export function SuccessToast({ children, onClose = () => {}, ...banner }) {
   )
 }
 
+export function InfoToast({ children, onClose = () => {}, ...banner }) {
```
dherault, line 66: We may want to put this in the DS.

dherault (approved):

.github/workflows/firebase-hosting-pull-request.yml
maciaszczykm: @dherault From Yarn 2 `yarn install --help`:

```
If the `--immutable` option is set (defaults to true on CI), Yarn will abort 
with an error exit code if the lockfile was to be modified (other paths can be 
added using the `immutablePatterns` configuration setting). For backward 
compatibility we offer an alias under the name of `--frozen-lockfile`, but it 
will be removed in a later release.
```

maciaszczykm: We could remove it there as it will be used by default in CI though.

.github/workflows/firebase-hosting-pull-request.yml
floreks: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on

This is a standard YAML syntax. YAML only requires quotes in a very limited number of scenarios, i.e. when string contains special characters.

.github/workflows/firebase-hosting-pull-request.yml
floreks: `CI=false` is there since create react app treats all warnings as errors in the CI and fails the build.
https://github.com/facebook/create-react-app/blob/f34d88e30c7d8be7181f728d1abc4fd8d5cd07d3/packages/react-error-overlay/build.js#L43-L51

Build output
```
Creating an optimized production build...
Compiled with warnings.

/static/js/main.eb3869c0.js is 7.97 MB, and won't be precached. Configure maximumFileSizeToCacheInBytes to change this limit.

Search for the keywords to learn more about each warning.
To ignore, add // eslint-disable-next-line to the line before.

File sizes after gzip:

  2.07 MB   build/static/js/main.eb3869c0.js
  36.18 kB  build/static/css/main.b91960b0.css

The bundle size is significantly larger than recommended.
Consider reducing it with code splitting: https://goo.gl/9VhYWB
You can also analyze the project dependencies: https://goo.gl/LeUzfb

The project was built assuming it is hosted at /.
You can control this with the homepage field in your package.json.

The build folder is ready to be deployed.
You may serve it with a static server:

  yarn global add serve
  serve -s build

Find out more about deployment here:

  https://cra.link/deployment
```

One way to solve that would be to introduce lazy loaded views and split the output bundle into separate chunks per view, but this is way more work, hence the `CI=false` solution.

www/src/components/utils/Toasts.js
floreks: The whole Toast component should be moved there, but since this was already a bigger change I only added this small wrapper method here.

floreks: I'll do that in a separate PR. I have already added Toast to the DS

PR #461, opened 2022-08-24
ENG-549 Account Refinements - Account attributes

dogmar:
Part 1 of updating all the Account pages. 
[Linear](https://linear.app/pluralsh/issue/ENG-549/account-refinement-overall-account-attributes)
[Linear parent issue](https://linear.app/pluralsh/issue/ENG-308/account-refinement)
[Figma](https://www.figma.com/file/iMEQMQJQTctPaLf8Lr44qu/Accounts-Tab?node-id=366%3A16842)

Screen Shot 2022-08-24 at 4 00 12 PM 1


michaeljguarino:
At looks like `Account Attributes` is the last in the list of tabs in the gutter, but it should actually be the first element

www/src/components/profile/Icon.tsx
```diff
@@ -0,0 +1,57 @@
...
+  lightest: 'fill-three-hover',
+}
+
+type IconProps = DivProps & {
+  icon: ReactNode
+  hue?: Hue
+  hover?: string
+}
+
+export const Icon = forwardRef<HTMLDivElement, IconProps>(({
```
floreks, line 21: This looks like a DS component candidate.

dogmar: For sure I can pull that out into the design system once this is all done.

dogmar:
> At looks like `Account Attributes` is the last in the list of tabs in the gutter, but it should actually be the first element

@michaeljguarino Looks like I bet on the wrong horse regarding this request: "Either default selection to account settings or switch order of tabs."
I've restored the tab order and now the page defaults to the "Account attributes" tab.

maciaszczykm (approved):
Looks great!

dherault (approved):
👍

PR #469, opened 2022-08-25
Use AppIcon and PageCard components

dogmar:
Updates to use new design system components `PageCard` and `AppIcon`. Updates old usage of `IconFrame` to new name of `AppIcon`.


michaeljguarino (approved):
looks like there are some lint issues but otherwise lgtm once fixed

PR #489, opened 2022-08-30
Update error handling for cloud shell

floreks:
## Summary

If I understand correctly, one way to forward error to the user is to do it directly from the `join` method. This way channel should return error and close if there is an error during initialization.

https://hexdocs.pm/phoenix/1.4.9/Phoenix.Channel.html#c:join/3

## Test Plan

Added unit test

## Checklist

### DIFF
--- a/apps/rtc/lib/rtc_web/channels/shell_channel.ex
+++ b/apps/rtc/lib/rtc_web/channels/shell_channel.ex
@@ -7,20 +7,15 @@ defmodule RtcWeb.ShellChannel do
   require Logger
 
   def join("shells:me", _, socket) do
-    send(self(), :connect)
-    {:ok, socket}
-  end
-
-  def handle_info(:connect, socket) do
     with %CloudShell{pod_name: name} = shell <- Shell.get_shell(socket.assigns.user.id),
          {:ok, _} <- Client.setup(shell),
          url <- Pods.PodExec.exec_url(name),
          {:ok, pid} <- Pods.PodExec.start_link(url, self()) do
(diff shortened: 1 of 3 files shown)


apps/rtc/lib/rtc_web/channels/shell_channel.ex
```diff
@@ -7,20 +7,15 @@ defmodule RtcWeb.ShellChannel do
...
          {:ok, _} <- Client.setup(shell),
          url <- Pods.PodExec.exec_url(name),
          {:ok, pid} <- Pods.PodExec.start_link(url, self()) do
-      {:noreply, assign(socket, :wss_pid, pid)}
+      {:ok, assign(socket, :wss_pid, pid)}
     else
       err ->
         Logger.info "failed to exec pod with #{inspect(err)}"
-        {:stop, {:shutdown, :failed_exec}, socket}
+        {:error, err}
```
michaeljguarino, line 18:
```elixir
Process.send_after(self(), :shutdown, 10 * 1000)
push(...)
{:ok, socket}
```

then add another handle_info like

```elixir
def handle_info(:shutdown, socket), do: {:stop, {:shutdown, :failed_boot}, socket}
```

michaeljguarino: actually thinking about this a bit more, the scale of our shell processes are so low, this optimization shouldn't really be necessary

floreks:
@michaeljguarino can you approve if it's ok?

michaeljguarino (approved):

PR #490, opened 2022-08-30
Implement Publisher screen

dherault:
image

To test the QA:
- Go to any repo.
- Click on "Plural" on the SideNav
Then:
- Create a publisher from the user menu
- Should redirect to the publisher screen

Also resolves ENG-576

## Checklist


nickcernera:
I added a QA ticket, [here](https://linear.app/pluralsh/issue/ENG-576/[qa]-publisher-view).

www/src/components/utils/IconUploadPreview.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </Flex>
+    )
+  }
+
+  return (
+    <Img
+      src={src}
+      alt="Icon"
+      width={64 + 32}
+      height={64 + 32}
```
maciaszczykm, line 48: Just asking but why not 96 as above?

www/src/components/publisher/CreatePublisherModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+        />
+      </FormField>
+    )
+  }
+
+  return (
+    <Modal
+      open={open}
+      onClose={onClose}
+      width={608}
```
maciaszczykm, line 160: You can use `size="large"` if you will sync with master.

maciaszczykm (commented):
Code looks good, I've left 2 tiny comments.

www/src/components/publisher/CreatePublisherModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+    const reader = generatePreview(file, (file: IconUploadType) => {
+      setIconUpload(file)
+    })
+
+    return () => {
+      reader.abort()
+    }
+  }, [iconPicker.files])
+
+  useEffect(() => {
```
michaeljguarino, line 87: this should be done in an `onCompleted` callback for the useMutation hook

www/src/components/publisher/CreatePublisherModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+      <Flex gap="medium">
+        {renderUrlField(
+          'Slack link', 'Slack invite URL', slack, setSlack, errors.slack
+        )}
+        {renderUrlField(
+          'Twitter link', 'Twitter URL', twitter, setTwitter, errors.twitter
+        )}
+      </Flex>
+      {!!preMutationError && (
+        <Alert severity="error">
```
michaeljguarino, line 258: why don't we use the ValidatedInput component for these?  Showing full alerts will be a pretty jarring ux

michaeljguarino: also it makes the code wayyy more verbose

dherault: I'm pushing toward the alert because the error is usually that the owner already has a publisher created. It does not fit any of the fields.

www/src/components/publisher/CreatePublisherModal.tsx
dherault: On it.

www/src/components/publisher/CreatePublisherModal.tsx
michaeljguarino: oh for errors from gql I think we should have an alert, for validation errors on the input fields themselves, something associated with the field makes more sense

michaeljguarino (approved):

maciaszczykm (approved):

PR #493, opened 2022-08-30
ENG-556 - Update marketplace token component

floreks:
## Summary

Updated styling of the token component.

image

## Test Plan

Locally

## Checklist


floreks:
@nickcernera should we use a chip or a token here?
Token by default has no border and uses font weight 400 I think
https://pluralsh-design.web.app/?path=/story/token--default

Standard chip has a border but font weight is 600 for the default size
https://pluralsh-design.web.app/?path=/story/chip--default&args=clickable:true

michaeljguarino (approved):

PR #497, opened 2022-08-30
ENG-308 - Account Refinement

dogmar:
Finishing up the remainder of the changes here: https://linear.app/pluralsh/issue/ENG-308/account-refinement


michaeljguarino:
The group modal with users and attributes all in the same view is a little strange currently.  I think the scroller needs a defined height to work, which is part of why it was nice to have its own tab rather than be placed in-line

michaeljguarino:
Also the modals across the board look like they probably need to be wider

www/src/components/account/CreateGroup.js
```diff
@@ -0,0 +1,88 @@
...
+    }),
+  })
+
+  return (
+    <>
+      <Button
+        secondary
+        onClick={() => setOpen(true)}
+      >
+        Create Group
```
maciaszczykm, line 40: In the designs it's not using title case: `Create group`.

www/src/components/account/CreateRole.js
```diff
@@ -0,0 +1,76 @@
...
+    onCompleted: () => resetAndClose(),
+  })
+
+  return (
+    <>
+      <Button
+        secondary
+        onClick={() => setOpen(true)}
+      >
+        Create Role
```
maciaszczykm, line 52: As above.

www/src/components/account/CreateGroup.js
nickcernera: +1—this is the only change required for VQA approval of this PR. Otherwise, everything is looking fantastic. Nice job, @dogmar.

nickcernera:
1) The modals definitely do need to be wider; I've already documented this in a [linear ticket](https://linear.app/pluralsh/issue/ENG-574/additional-modal-size-variant).

2) I don't think that I put enough thought into the functionality of the whole adding/searching users functionality on this modal. As such, I think it might be wise to temporarily replace the entire user component with the "Add users as chips" component. I will also create a backlogged design ticket to give this extra attention once resources open up. @michaeljguarino @dogmar are you happy with this plan?

3) There seems to have been a change in the mockup that was an experimentation and shouldn't have been in the Delivery Specs. That's my bad. Could you also revert this component to the normal `FormField` component?
CleanShot 2022-08-31 at 11 53 45@2x

michaeljguarino:
> > 2. I don't think that I put enough thought into the functionality of the whole adding/searching users functionality on this modal. As such, I think it might be wise to temporarily replace the entire user component with the "Add users as chips" component. I will also create a backlogged design ticket to give this extra attention once resources open up. @michaeljguarino @dogmar are you happy with this plan?
> 
> After I started to implement this, I realized that we need to lazy-load the users in a group since there could be hundreds of users. The chip list isn't built to handle that. I think it might be best to go back to the original two-tabs design until we've got a better solution.

Agreed, I think the two-tabs solution is probably going to be best here.

dogmar:
All requested changes are in. PTAL @michaeljguarino @maciaszczykm @nickcernera

michaeljguarino (approved):

nickcernera:
That makes sense. Go ahead and implement it that way imo.

PR #512, opened 2022-09-08
Fix user search query

michaeljguarino:
## Summary

It currently doesn't support case insensitivity and only works for prefixes.  This should help a bit with that.

## Test Plan
existing tests

## Checklist


dogmar:
Could it also support searching the middle of the string as well, so you can find by last name and such?

michaeljguarino:
Oh yeah it should do that with this change (adding wildcards to both ends)

floreks (approved):

maciaszczykm (approved):

PR #537, opened 2022-09-19
Add validation for repository names

michaeljguarino:
## Summary
Repo names need to be valid k8s namespaces, but we need to enforce this in the api.  This regex should do a good enough job of that.

## Test Plan
fix unit tests

## Checklist

### DIFF
--- a/apps/core/lib/core/schema/repository.ex
+++ b/apps/core/lib/core/schema/repository.ex
@@ -198,7 +198,7 @@ defmodule Core.Schema.Repository do
     |> cast_assoc(:shell)
     |> foreign_key_constraint(:publisher_id)
     |> cast_assoc(:integration_resource_definition)
-    |> validate_format(:name, ~r/^[a-z][-a-z0-9]*$/, message: "name must be formatted as a valid kubernetes namespace")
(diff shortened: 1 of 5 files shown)


maciaszczykm (approved):

floreks:
Namespace regex follows the RFC 1123 standard to match the name. Names like `1-ab3` are also valid. This is the official regex used by the kubectl to validate the namespace name: `[a-z0-9]([-a-z0-9]*[a-z0-9])?`.

apps/core/lib/core/schema/repository.ex
```diff
@@ -198,6 +198,7 @@ defmodule Core.Schema.Repository do
     |> cast_assoc(:shell)
     |> foreign_key_constraint(:publisher_id)
     |> cast_assoc(:integration_resource_definition)
+    |> validate_format(:name, ~r/^[a-z][-a-z0-9]*$/, message: "name must be formatted as a valid kubernetes namespace")
```
floreks, line 201: Namespace regex follows the RFC 1123 standard to match the name. Names like `1-ab3` are also valid. This is the official regex used by the kubectl to validate the namespace name: `[a-z0-9]([-a-z0-9]*[a-z0-9])?`.

michaeljguarino: Ah cool, will change that

PR #542, opened 2022-09-19
[ENG-303] Improve error message for provider query

floreks:
## Summary

Updated terraform provider error message when it is not found.

[Linear](https://linear.app/pluralsh/issue/ENG-303/querying-an-unknown-provider-version-returns-an-useless-error)

## Test Plan

Added unit test

## Checklist

### DIFF
--- a/apps/core/lib/core/schema/provider_scaffold.ex
+++ b/apps/core/lib/core/schema/provider_scaffold.ex
@@ -5,23 +5,23 @@ defmodule Core.Schema.ProviderScaffold do
   def available(), do: @providers
 
   def new(name, vsn \\ :default) when name in @providers do
-    %__MODULE__{
-      name: name,
-      content: provider_file(name, vsn)
-    }
+    with {:ok, content} <- provider_file(name, vsn) do
+      {:ok, %__MODULE__{name: name, content: content}}
+    end
   end
 
   defp provider_file(name, :default) do
     Path.join([:code.priv_dir(:core), "scaffolds/terraform/providers", "#{name}.eex"])
-    |> File.read!
+    |> File.read()
+    |> handle_read(name, :default)
   end
 
   defp provider_file(name, vsn) do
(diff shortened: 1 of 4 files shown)


maciaszczykm (approved):

maciaszczykm (approved):

apps/core/lib/core/schema/provider_scaffold.ex
```diff
@@ -18,6 +18,9 @@ defmodule Core.Schema.ProviderScaffold do
 
   defp provider_file(name, vsn) do
     Path.join([:code.priv_dir(:core), "scaffolds/terraform/providers/#{vsn}", "#{name}.eex"])
-    |> File.read!
+    |> case do
+       {:ok, resp} -> {:ok, resp} |> File.read!
```
michaeljguarino, line 22: Path.join returns an ok tuple? also it's probably cleaner to write

```elixir
File.read!({:ok, resp})
```
Since it's inline like this and only one func call

floreks:
@michaeljguarino PTAL

apps/core/lib/core/schema/provider_scaffold.ex
```diff
@@ -18,6 +18,9 @@ defmodule Core.Schema.ProviderScaffold do
 
   defp provider_file(name, vsn) do
     Path.join([:code.priv_dir(:core), "scaffolds/terraform/providers/#{vsn}", "#{name}.eex"])
-    |> File.read!
+    |> File.read
+    |> case do
```
michaeljguarino, line 22: You need to also handle the `{:ok, result}` case, probably with a case clause like:

```elixir
{:ok, contents} -> contents
```

floreks:
Done PTAL

michaeljguarino (commented):
This is still a little wonky, what it should do is return an `:ok`/`:error` tuple and bubble that up to the resolver where appropriate.  I can take on the branch and show you a bit how it should look

floreks:
@michaeljguarino thanks! All clear now.

PR #545, opened 2022-09-19
Onboarding applications selection

dherault:
## Summary

Adds a new screen before the onboarding starts, right before we choose a git provider.

Left to do it the consecutive PRs:
- Use localStorage to store the ids/names of the selected apps
- Make restart link operational

image

## Checklist


www/src/components/shell/onboarding/applications/ApplicationsSelection.tsx
```diff
@@ -0,0 +1,156 @@
...
+        gridTemplateRows="repeat(auto-fill, 42px)"
+        gridColumnGap="16px"
+        gridRowGap="16px"
+        paddingRight="xsmall"
+        paddingBottom="medium"
+      >
+        {filteredApplications.map(application => (
+          <RepositoryChip
+            imageUrl={application.icon}
+            label={capitalize(application.name)}
```
maciaszczykm, line 123: It will look a bit strange for some names like `csgo` etc. I'd vote for consistency here and try to make it look the same in every view.

maciaszczykm (commented):
Looks good. I've two tiny comments. One is in the code, second one is that while my cloud shell was loading vertical stepper had no spacing from the left side. It was directly on `stacks` branch.

www/src/components/shell/onboarding/applications/ApplicationsSelection.tsx
dogmar: @Agreed, but otherwise, LGTM.

www/src/components/shell/queries.js
```diff
@@ -61,3 +62,22 @@ export const POLL_DEMO_PROJECT_QUERY = gql`
...
+  query ApplicationsQuery($cursor: String) {
+    repositories(after: $cursor, first: 200) {
+      pageInfo {
+        ...PageInfo
+      }
+      edges {
+        node {
+          id
+          name
+          icon
```
michaeljguarino, line 76: we'll need dark icon too

michaeljguarino (commented):
We're also going to have to have some way to tie in the relevant bundles here.  Also I suspect we could actually have a hardcoded list instead of returning all applications since only certain ones are relevant in onboarding in all likelihood

floreks:
Should there be an option to continue anyway without selecting an application?

dherault:
I think the skip demo button takes you directly to the shell

www/src/components/shell/onboarding/applications/ApplicationsSelection.tsx
dherault: The design ask for it.

www/src/components/shell/queries.js
dherault: Done

dogmar (approved):

maciaszczykm (approved):

PR #546, opened 2022-09-20
Add install panel for stacks

maciaszczykm:
## Summary
Zrzut ekranu 2022-09-20 o 11 19 23
Zrzut ekranu 2022-09-20 o 11 19 07

Adds install dropdown panel for stacks.

Note: Panel has the same width as in the repository details view. I was not able to stretch it to 200px.

Closes ENG-617.

## Test Plan
Tested manually.

## Checklist


www/src/components/stack/misc.tsx
```diff
@@ -0,0 +1,19 @@
+import InstallDropdownButton, { Recipe } from 'components/utils/InstallDropdownButton'
+
+import { StackContext } from './types'
+
+export function StackActions({ stack }: StackContext) {
+  const recipes = stack?.collections?.map(({ name, provider }) => ({
+    name,
+    description: `Installs ${stack.name} on ${provider}`,
+    provider,
+  } as Recipe))
```
maciaszczykm, line 10: A bit of workaround to avoid duplicating `InstallDropdownButton` code.

dogmar (approved):

www/src/components/utils/InstallDropdownButton.tsx
```diff
@@ -8,9 +8,35 @@ import {
 import { Link } from 'react-router-dom'
 import capitalize from 'lodash/capitalize'
 
-import RepositoryContext from '../../contexts/RepositoryContext'
+export interface Recipe {
```
floreks, line 11: Since the GraphQL generated types were merged already we can use them instead of creating duplicated types (src/generated/graphql.tsx).

maciaszczykm: This is only a common pary of repository and stack recipes (data that is required to render the component). In this case I would like to keep it minimal instead of using generated type.

PR #582, opened 2022-09-29
ENG-732 - Resolve "install plural console" issues with onboarding flow.

dogmar:
# Description
Fixes the following ([linear ticket](https://linear.app/pluralsh/issue/ENG-732/resolve-install-plural-console-issues-with-onboarding-flow)):
- The app selection screen should include a toggle switch for "Install Plural Console"
- During a quickstack install with Plural Console selected, the command is not installing the Plural Console.
- During a pre-made stack install, with Plural Console selected, there should be an additional step to install plural console (prior to stack install)

# Testing
Manually tested flow to terminal launch and install step for both quick-stack and pre-made stack once each.

### Showing 'console' is selected for install
Screen Shot 2022-09-29 at 12 56 32 PM

### DIFF
--- a/www/src/components/shell/onboarding/applications/ApplicationsSelection.tsx
+++ b/www/src/components/shell/onboarding/applications/ApplicationsSelection.tsx
@@ -2,6 +2,7 @@ import {
   useCallback,
   useContext,
   useEffect,
+  useMemo,
   useState,
 } from 'react'
 import {
@@ -50,6 +51,8 @@ const hueToColor = {
   red: 'text-error-light',
 }
 
+const CONSOLE_APP_NAME = 'console'
+
 const searchOptions = {
   keys: ['name'],
   threshold: 0.25,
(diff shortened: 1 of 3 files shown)


www/src/components/shell/onboarding/applications/ApplicationsSelection.tsx
```diff
@@ -325,10 +369,24 @@ function ApplicationsSelection({ onNext }: ApplicationsSelectionProps) {
...
+              checked={
+                !!selectedApplications.find(a => a.id === application.id)
+              }
               onClick={() => toggleApplication(application)}
-              cursor={isStack ? 'auto' : selectedApplications.length >= MAX_SELECTED_APPLICATIONS && !selectedApplications.find(a => a.id === application.id) ? 'not-allowed' : 'pointer'}
-              opacity={selectedApplications.length >= MAX_SELECTED_APPLICATIONS && !selectedApplications.find(a => a.id === application.id) ? 0.5 : 1}
+              cursor={
+                isStack
+                  ? 'auto'
+                  : countSelectedApps(selectedApplications) >= MAX_SELECTED_APPLICATIONS
```
michaeljguarino, line 379: could we save this as a boolean above so it doesn't have to be computed on each invocation?

dogmar: Done.

michaeljguarino (approved):

PR #584, opened 2022-09-30
Fix broken Save button on Edit Repository pages

dogmar:
The new community urls weren’t getting saved either.

### DIFF
--- a/www/src/components/repository/Repository.js
+++ b/www/src/components/repository/Repository.js
@@ -74,7 +74,7 @@ function Repository() {
           <ResponsiveLayoutSpacer />
           <TabPanel
             as={
-              <ResponsiveLayoutContentContainer />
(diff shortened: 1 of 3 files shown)


floreks:
This is from our API schema
image

You can check all available fields here: https://app.plural.sh/graphiql

michaeljguarino (commented):
Are you able to test it?  I can give you the permissions you need to do so if you aren't

dogmar:
@michaeljguarino Oh, I'm able to test and everything writes/reads from these fields just fine. I just wasn't sure if they were the right fields to be mapping to. Like maybe `documentation` was meant for actual docs text and not just a url? And `gitUrl` doesn't sound github specific. Also inconsistent that it's the only one that actually has `url` in its name.

dherault (approved):

michaeljguarino:
I see, it looks good to go from what i can tell

PR #587, opened 2022-10-03
[ENG-725] Detect detached message and trigger shell restart

floreks:
## Summary

It appears that the workspace detached message does not really close the WS connection and it is still considered alive. Heartbeat can be seen every 30s in the network panel. Since there is no real connection error, timeout changes would not impact anything here. They are actually quite short (10s/20s).

I have added simple logic to detect the detached message and trigger the shell restart.

@michaeljguarino @dherault WDYT?

[Nagranie ekranu z 04.10.2022 14:44:51.webm](https://user-images.githubusercontent.com/2285385/193822877-13ae3a38-ff84-4f30-84c4-ccef49244979.webm)

## Test Plan

Locally with 2 tabs. Opening shell in a second tab actually results in the detached from workspace message.

## Checklist

### DIFF
--- a/www/src/components/shell/Shell.js
+++ b/www/src/components/shell/Shell.js
@@ -59,7 +59,7 @@ function Shell({ shell }) {
     chan.on('stdo', ({ message }) => {
       const decoded = decodeBase64(message)
 
-      term.write(decodeBase64(message))
+      term.write(decoded)
 
       if (!restart && decoded.includes(detachedMessage)) {
         setRestart(true)
--- a/www/src/components/shell/Shell.js
+++ b/www/src/components/shell/Shell.js
@@ -28,6 +28,7 @@ import useOnboarded from './onboarding/useOnboarded'
 const { Buffer } = require('buffer/')
 
 const decodeBase64 = str => Buffer.from(str, 'base64').toString('utf-8')
+const detachedMessage = '[detached (from session workspace)]'
 
 function Shell({ shell }) {
   const xterm = useRef(null)
@@ -37,9 +38,11 @@ function Shell({ shell }) {
   const fitAddon = useMemo(() => new FitAddon(), [])
   const [terminalTheme] = useContext(TerminalThemeContext)
   const { fresh } = useOnboarded()
+  const [restart, setRestart] = useState(false)
 
   useEffect(() => {
     if (!xterm?.current?.terminal) return
+    if (restart) return
 
     const term = xterm.current.terminal
     const chan = socket.channel('shells:me')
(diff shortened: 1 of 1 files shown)


www/src/components/shell/Shell.js
```diff
@@ -37,9 +38,11 @@ function Shell({ shell }) {
   const fitAddon = useMemo(() => new FitAddon(), [])
   const [terminalTheme] = useContext(TerminalThemeContext)
   const { fresh } = useOnboarded()
+  const [restart, setRestart] = useState(false)
 
   useEffect(() => {
     if (!xterm?.current?.terminal) return
+    if (restart) return
```
michaeljguarino, line 45: does this actually restart the terminal?

www/src/components/shell/Shell.js
```diff
@@ -52,8 +55,16 @@ function Shell({ shell }) {
     }
 
     term.write(`Booting into your ${shell.provider} shell...\r\n\r\nIt can take a few minutes to load. Try refreshing the page if it gets stuck for too long.\r\n`)
-    chan.onError(err => console.error(`Unknown error during booting into your shell: ${err}`))
-    chan.on('stdo', ({ message }) => term.write(decodeBase64(message)))
+    chan.onError(err => console.error(`Unknown error during booting into your shell: ${JSON.stringify(err)}`))
+    chan.on('stdo', ({ message }) => {
+      const decoded = decodeBase64(message)
+
+      term.write(decodeBase64(message))
```
michaeljguarino, line 62: is there a reason we don't pass the decoded const here instead of doube b64 decoding?

dherault:
I'll let Michael handle this one.

www/src/components/shell/Shell.js
floreks: Definitely no reason. I forgot to update this line. Good catch :slightly_smiling_face:

www/src/components/shell/Shell.js
floreks: Actually, this line is so that this effect won't run twice in a row since `restart` is first set to true by this effect and then to false by another one a bit below. This way "restart" will actually happen only once.

floreks: If you know a good way to trigger this only once, maybe using some other react feature, I'd love to do it differently. This is a bit of a workaround.

floreks:
@michaeljguarino PTAL

www/src/components/shell/Shell.js
```diff
@@ -37,9 +38,11 @@ function Shell({ shell }) {
   const fitAddon = useMemo(() => new FitAddon(), [])
   const [terminalTheme] = useContext(TerminalThemeContext)
   const { fresh } = useOnboarded()
+  const [restart, setRestart] = useState(false)
 
   useEffect(() => {
     if (!xterm?.current?.terminal) return
```
michaeljguarino, line 44: change this clause to:

```js
if (!xterm?.current?.terminal || restart) {
  restart && setRestart(false)
  return
}
```

and remove the second hook.  I think the logic will be clearer that way

floreks: oh yeah, that's definitely easier to understand :slightly_smiling_face:

michaeljguarino (approved):

PR #594, opened 2022-10-04
[ENG-721] Add history router to better handle state of redirect

floreks:
## Summary

- Added a wrapper for the original router to be able to keep track of view changes
- Exposed a custom `useHistory` hook that can be used to navigate to the last view or the default one if no last view is found
- Last visited "functional" views (except related to login) are saved and stored in the local storage. 

**Important:** This feature will basically always redirect existing users to the last seen view before logging out or to the desired view in case the user tried to enter some page and was redirected to the login view instead.

**Note:** History is tied to a browser, not the user. It means that after logging out and logging in user will be redirected to the last accessed view, no matter the account. We could easily mitigate that by clearing the local storage when the user logs out manually.

## Test Plan

Locally using different logging methods:
- via Google/GitHub provider
- Username/Password
- email on `plural.sh` domain (passwordless)
- providing non-existing email and creating a new account using the signup process

## Checklist


michaeljguarino (approved):

michaeljguarino:
There used to be a useHistory hook in react-router, is it not available anymore?

floreks:
It has been removed in the react-router v6. That was the first thing I considered. Unfortunately, it wouldn't help us much since it doesn't offer any kind of filtering that we could use to omit i.e. login paths and store the state inside the local storage.

dherault (approved):

PR #615, opened 2022-10-13
Azure Cloud Shell

dogmar:


www/src/components/shell/onboarding/cloud/CloudCredentials.js
```diff
@@ -42,44 +46,51 @@ function CloudCredentials() {
...
-        )}
-        {exceptions && <Exceptions exceptions={exceptions} />}
+            <Select
+              width="100%"
+              onChange={({ target: { value } }) => {
+                setProvider(value)
+              }}
+              value={provider}
+            >
+              {CLOUDS.map(cloud => (
```
michaeljguarino, line 65: Could we add a growthbook feature flag to control if azure is included in this list?

michaeljguarino: Actually wait I'll do it since I have growthbook access

michaeljguarino (approved):

www/src/components/shell/onboarding/cloud/azure.tsx
```diff
@@ -0,0 +1,195 @@
...
+            <MenuItem
+              key={region}
+              value={region}
+            >
+              {region}
+            </MenuItem>
+          ))}
+        </Select>
+      </FormField>
+      <FormField
```
michaeljguarino, line 128: We should probably add the validations here like in other forms as well

www/src/components/shell/onboarding/cloud/azure.tsx
```diff
@@ -0,0 +1,195 @@
...
+  'japaneast',
+  'koreacentral',
+  'canadacentral',
+  'francecentral',
+  'germanywestcentral',
+  'norwayeast',
+  'brazilsouth',
+]
+
+export const AZURE_VALIDATIONS = [
```
michaeljguarino, line 32: Is this being used?

www/src/components/shell/onboarding/cloud/azure.tsx
```diff
@@ -0,0 +1,195 @@
...
+    name: 'region',
+  },
+  {
+    field: 'credentials.workspace.project',
+    func: stringExists,
+    name: 'project',
+  },
+  {
+    field: 'credentials.azure.clientId',
+    func: stringExists,
```
michaeljguarino, line 45: I'm pretty sure client id, tenant id, and subscription are all standard uuids, so we could validate that to prevent some misconfig

michaeljguarino: here's a regex that matches uuids if we don't have it already in the codebase: `^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`

michaeljguarino (approved):

PR #642, opened 2022-10-25
[ENG-938] - Fix issues with domain list/detail view and add option to filter domains

floreks:
## Summary

- Fixed an issue with invisible delete icon button (fixes ENG-904 also)
- Fixed back button
- Added simple filter option (frontend only) to filter the domains (it can be improved later to use the API)

image
image

## Test Plan

Locally

## Checklist

### DIFF
--- a/www/src/components/account/Domains.js
+++ b/www/src/components/account/Domains.js
@@ -2,34 +2,43 @@ import { useMutation, useQuery } from '@apollo/client'
 import { Box } from 'grommet'
 import { Avatar, Flex, Span } from 'honorable'
 import moment from 'moment'
-import { ListBoxItem, Modal, PageTitle } from 'pluralsh-design-system'
-import { useMemo, useState } from 'react'
+import {
+  ListBoxItem, Modal, PageTitle, SearchIcon,
+} from 'pluralsh-design-system'
+import { useCallback, useMemo, useState } from 'react'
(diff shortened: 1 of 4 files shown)


www/src/components/account/Domains.js
```diff
@@ -198,20 +207,25 @@ function Domain({ node, last, setDomain }) {
   )
 }
 
-export function Domains() {
+function DomainsInner({ q, setDomainSelected }) {
   const [listRef, setListRef] = useState(null)
   const [domain, setDomain] = useState(null)
   const { data, loading, fetchMore } = useQuery(DNS_DOMAINS, {
```
michaeljguarino, line 213: We should modify the api to allow for a search filter and not do it in the frontend (it's ok for now though)

floreks: Totally agree. This is just a quick fix since it is really useful to be able to filter at all. Will create a follow-up issue to add support in the API and update the frontend.

michaeljguarino (approved):

www/src/components/account/Domains.js
michaeljguarino: Cool sounds good

dogmar (approved):

maciaszczykm (approved):

PR #683, opened 2022-11-08
[ENG-940] - Refresh user on onboarding status update

floreks:
## Summary

Trigger user update when user onboarding status is updated. It was using old cached version of user object, that's why user could still see the skip demo button after getting back to the shell view.

It adds a `setRefresh` function to the user context value to allow to immediately trigger the refresh after any action. It might be useful in many cases.

## Test Plan

Locally

## Checklist


www/src/components/shell/onboarding/useOnboarded.ts
```diff
@@ -9,10 +9,11 @@ import { UPDATE_USER } from '../../users/queries'
 const FORCE_ONBOARDING = 'plural-force-onboarding'
 
 function useOnboarded() {
-  const me = useContext(CurrentUserContext)
+  const { me, setRefresh } = useContext(CurrentUserContext)
 
   const [mutation] = useMutation(UPDATE_USER, {
     variables: { attributes: { onboarding: OnboardingStatus.ONBOARDED } },
+    onCompleted: () => setRefresh(true),
```
michaeljguarino, line 16: is this not done automatically (also you can just pass the refetch function itself through the context right? instead of using the hook to do it for you)

michaeljguarino: for context, apollo's cache should auto-update after mutations, perhaps the mutation needs to query some more fields or something

floreks: Cache is not getting updated unfortunately. This might be a similar issue to what @maciaszczykm encountered when working on stacks loading? @maciaszczykm do you remember the exact issue there? I know it was related to cache and some fragment naming or something like that.

floreks: > is this not done automatically (also you can just pass the refetch function itself through the context right? instead of using the hook to do it for you)

sure, I will pass it directly

www/src/components/shell/onboarding/useOnboarded.ts
```diff
@@ -5,14 +5,26 @@ import { useLocation } from 'react-router-dom'
...
+    update: (cache, { data }) => {
+      cache.modify({
+        fields: {
+          me: () => cache.writeFragment({
+            data: data?.updateUser,
+            fragment: UserFragment,
+          }),
+        },
+      })
+    },
```
floreks, line 27: FYI @maciaszczykm @michaeljguarino 

When a mutation happens, apollo does not update the fragment cache automatically. They suggest to use a refetch policy
```ts
  refetchQueries: [
    {query: GET_POST}, // DocumentNode object parsed with gql
    'GetComments' // Query name
  ],
```
or update the cache manually as I did here. 

Ref: https://www.apollographql.com/docs/react/data/mutations/#updating-local-data

floreks:
PTAL. I was able to avoid calling the API and updated the cache manually.

maciaszczykm (approved):

PR #706, opened 2022-11-10
move plural workers to their own node group

DavidSpek:
## Summary
Move Plural workers onto their own nodes so that they don't interfere with other parts of Plural.

## Checklist

### DIFF
--- a/plural/terraform/aws/variables.tf
+++ b/plural/terraform/aws/variables.tf
@@ -86,7 +86,7 @@ variable "single_az_node_groups" {
     plural_small = {
       name = "plural-small"
       capacity_type = "ON_DEMAND"
-      min_capacity = 1
+      min_capacity = 3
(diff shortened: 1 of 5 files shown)


plural/terraform/aws/variables.tf
```diff
@@ -103,6 +103,26 @@ variable "single_az_node_groups" {
         }
       ]
     }
+    plural_worker_small = {
+      name = "plural-worker-small"
+      capacity_type = "ON_DEMAND"
+      min_capacity = 0
```
michaeljguarino, line 109: we should have at least 1 as min capacity (there will always be some workers)

michaeljguarino: (we should do this for the other group as well)

DavidSpek: done

PR #707, opened 2022-11-10
Move docker scan into GenStage

michaeljguarino:
## Summary

This will allow full control on scan parallelism.  The current setup should guarantee only one scan is run concurrently, and we can scale up as needed.

We've seen bursts of scans overburden nodes so need more fine grained control here

## Test Plan
modified unit tests

## Checklist

### DIFF
--- a/apps/worker/lib/worker/docker/pipeline.ex
+++ b/apps/worker/lib/worker/docker/pipeline.ex
@@ -3,14 +3,12 @@ defmodule Worker.Docker.Pipeline do
   require Logger
 
   def start_link(producer) do
-    Flow.from_stages([producer], stages: 1, max_demand: 10)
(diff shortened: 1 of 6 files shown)


apps/worker/lib/worker/docker/pipeline.ex
```diff
@@ -3,14 +3,12 @@ defmodule Worker.Docker.Pipeline do
   require Logger
 
   def start_link(producer) do
-    Flow.from_stages([producer], stages: 1, max_demand: 10)
+    Flow.from_stages([producer], stages: 1, max_demand: 5)
```
DavidSpek, line 6: Does it make sense to have this configurable through an env var as well?

michaeljguarino: I can make that change

PR #709, opened 2022-11-10
Refactor our polled workers to use a common macro

michaeljguarino:
## Summary
This should help keep the code in this pattern relatively consistent and reusable.  Also completely removed the rabbit based scan logic.

## Test Plan
unit tests

## Checklist


plural/helm/plural/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: plural
 description: A helm chart for installing plural
-appVersion: "0.8.9"
-version: 0.8.91
+appVersion: "0.8.10"
+version: 0.8.92
```
DavidSpek, line 5: It seems like this is outdated and needs to be rebased.

PR #719, opened 2022-11-11
fix(deps): upgrade to Elixir 1.12.3 and Erlang 24.3.4.6

DavidSpek:
## Summary
This PR updates our backend to the next version of Elixir and Erlang. This has been made possible by a number of dependency updates and a few small fixes included in this PR.

Blocked by:
- ~https://github.com/pluralsh/plural/pull/720~
- ~https://github.com/pluralsh/plural/pull/721~
- ~https://github.com/pluralsh/plural/pull/714~

## Test Plan
Built images and deployed on the dev cluster. No obvious errors in the logs. Everything seems to be working as expected.

## Checklist


.github/workflows/test.yaml
```diff
@@ -87,18 +87,18 @@ jobs:
...
-          path: deps
-          key: ${{ runner.os }}-mix-5-${{ hashFiles('**/mix.lock') }}
-          restore-keys: ${{ runner.os }}-mix-5
-      - name: Restore _build
-        uses: actions/cache@v3
-        with:
-          path: _build
-          key: ${{ runner.os }}-mix-5-${{ hashFiles('**/mix.lock') }}
-          restore-keys: ${{ runner.os }}-mix-5
+      # - name: Restore dependencies cache
```
michaeljguarino, line 90: we should keep the dependencies cache for sure, it likely just needs the cache key to be bumped

DavidSpek: Yeah that is why I disabled for now, since the PR was just being used for the tests up to this point. We could probably also be a bit smarter in terms of the key naming, like if we use the elixir and/or erlang versions instead we could build/test against multiple version with cache more easily and it would be clear what cache key to use.

michaeljguarino (approved):

michaeljguarino:
the most likely possible broken aspect of this would be the cloud shell, because of the change to ssh key generation, but I'm pretty sure it's ok and did sanity check

PR #736, opened 2022-11-15
Paginate docker images in ImageVersionPicker on Docker page

dherault:
## Summary

Adds pagination for this component:
image

## Checklist


www/src/components/repos/Docker.tsx
```diff
@@ -107,14 +107,27 @@ function ImageVersionPicker({ image }: any) {
...
+  const { data } = useQuery(DOCKER_IMG_Q, {
+    variables: {
+      dockerRepositoryId: dockerRepository.id,
+      cursor,
+    },
   })
 
-  if (!data || loading) return null
+  useEffect(() => {
+    if (data?.dockerImages) {
```
michaeljguarino, line 120: Is this just going to fetch the entire set of images until it stops?  We should only do it when a user scrolls into view for sure, some of these repos have thousands of images

maciaszczykm: I think in designs for this there was a "view more" button.

dherault:
I'm very unhappy with the end result.

Do to the implementation of the new Select, two problems arise:
- The select closes on fetch more click
- `// @ts-nocheck` is needed at the top of the file because `Select` admits only `ReactElement<ListBoxItemBaseProps> | ReactElement<ListBoxItemBaseProps>[]` as children. 

Moreover when clicking on a paginated version, the page reloads and the pagination resets, leaving the version missing from the select.

I say we revert to my previous "all fetch" implementation.

maciaszczykm:
> The select closes on fetch more click

Would `e.stopPropagation()` help with this as temp workaround?

www/src/components/repos/Docker.tsx
```diff
@@ -157,6 +184,17 @@ function ImageVersionPicker({ image }: any) {
             )}
           />
         ))}
+        {hasNextPage ? (
```
maciaszczykm, line 187: Would `{hasNextPage && (` work here instead?

maciaszczykm: Actually, this should be moved to `dropdownFooter` prop. See the example from Storybook:

```
dropdownFooter={
  shownLimit < items.length && (
    <ListBoxFooterPlus>View more</ListBoxFooterPlus>
  )
}
```

maciaszczykm: Dropdown closing on click will also stop being a problem when this prop will be used.

dherault: No... Because of problem number two

maciaszczykm: It was my guess as it works like this here: https://pluralsh-design.web.app/?path=/story/select--default (see "Click me!" button).

dherault: OH that's great!

www/src/components/repos/Docker.tsx
```diff
@@ -1,4 +1,7 @@
+// HACK until Select is fixed to admit any ReactNode
+// @ts-nocheck
```
maciaszczykm, line 2: Is this needed?

maciaszczykm (approved):
LGTM, one question added.

PR #757, opened 2022-11-17
Fix service worker registration

floreks:
## Summary

Vite handles the service worker a bit differently than CRA. This PR updates our configuration to make it work again.

## Test Plan

Locally with prod build.

## Checklist


maciaszczykm (approved):

www/src/serviceWorkerRegistration.ts
```diff
@@ -1,7 +1,7 @@
 const isProduction = import.meta.env.PROD
-const url = import.meta.env.BASE_URL || ''
+const url = import.meta.url || ''
```
michaeljguarino, line 2: is url unused?

floreks: It is used in the register method below.

DavidSpek (approved):
Deployed and all works as expected. No errors with registering the service worker anymore.

PR #768, opened 2022-11-17
Add secret command to input GitHub/GitLab token on onboarding

dherault:
## Summary

So, 
This PR create a secret token management flow for the onboarding.

How to use:
1 - Go to app.plural.sh and start the onboarding, after you've clicked on GitHub/GitLab and been redirected, 
type `t o k e n o u t`. 
The token should appear in an old-fashioned alert.
2 - Go to your dev/staging env. 
At the start of the onboarding type `t o k e n i n` 
(be careful of the applications search bar input autofocus) 
and input your token in the appearing prompt.
3 - You can now go through the onboarding with your updated token.
4 - Remember: the token expires after some time.

Also:
- The file `debug-tokens` is gone.
- I fixed a react keys warning on Router

## Checklist

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -84,6 +84,7 @@
     "react-icons": "4.6.0",
     "react-json-view": "1.21.3",
     "react-multiline-clamp": "2.0.0",
+    "react-piano-keys": "^1.2.1",
     "react-portal": "4.2.2",
     "react-resize-detector": "7.1.2",
     "react-router-dom": "6.4.3",
--- a/www/package.json
+++ b/www/package.json
@@ -84,7 +84,7 @@
     "react-icons": "4.6.0",
     "react-json-view": "1.21.3",
     "react-multiline-clamp": "2.0.0",
-    "react-piano-keys": "^1.2.1",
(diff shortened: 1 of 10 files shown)


michaeljguarino (commented):
Could we make the key sequence more memorable (it's not like it's actually a secret).  Perhaps something like "g i t t o k e n"

dherault:
We need two sequences, what do you suggest?

floreks:
We actually don't need 2 sequences. It is easier to generate token directly on the GitHub page. Token can be granted only specific permissions. We could also maybe use growthbook to enable this workaround for us instead of having to remember the sequence.

michaeljguarino:
It couldn't be like `gittoken out` and `gittoken in` as the sequences?  I don't think the one you chose is that hard to remember either, if we were to use it

dherault:
New sequences in description.

dherault:
I still think we need two sequences because it may be hard for the design team to go to github and issue a token. Hard but feasible ofc.

maciaszczykm (approved):

PR #774, opened 2022-11-18
chore(deps): upgrade hydra chart

DavidSpek:
## Summary
PR that will be used to update Plural too the latest release.

## Checklist


plural/helm/plural/values.yaml.tpl
```diff
@@ -95,7 +90,9 @@ hydra:
     public:
       hosts:
       - host: {{ $hydraHost }}
-        paths: ["/.*"]
+        paths:
+        - path: /
```
michaeljguarino, line 94: This probably needs to be `/.*` still if we've enabled regex for that ingress

DavidSpek: I think this is covered by `pathType: ImplementationSpecific`.

michaeljguarino (approved):
we can merge this but we should test oidc logins immediately after to be prepared to revert

PR #786, opened 2022-11-23
fix(deps): Update regex

floreks:
## Summary

Another attempt to fix the regex :slightly_smiling_face: I didn't test on the whole file earlier, just one line. Now it should be fine.

image

## Test Plan

regex101

## Checklist


DavidSpek:
What regex engine do you have selected on the left? Renovate is javascript based, but the regex in renovate can be a bit strange sometimes.

floreks:
I have checked JS also. I haven't noticed any major differences between JS and PHP (the default one).

michaeljguarino (approved):

PR #788, opened 2022-11-23
Fix xterm-addon-fit

michaeljguarino:
## Summary

we upgraded this dependency to an apparently broken version.  Downgrading allows us to properly resize the window again, but I'm seeing some odd screen interactions I'll likely still need to debug

## Test Plan
local

## Checklist

### DIFF
--- a/www/src/components/shell/Shell.tsx
+++ b/www/src/components/shell/Shell.tsx
@@ -71,7 +71,6 @@ function Shell({ shell }: any) {
     setChannel(chan)
 
     return () => {
-      socket.off([ref])
       chan.leave()
     }
   }, [shell, xterm, restart])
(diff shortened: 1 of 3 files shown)


www/src/components/shell/Shell.tsx
```diff
@@ -71,28 +68,36 @@ function Shell({ shell }: any) {
       // }
     })
     chan.join()
+    setChannel(chan)
+
+    return () => {
+      socket.off([ref])
+      chan.leave()
+    }
+  }, [shell, xterm, restart])
```
dogmar, line 77: Wanna add that missing `ref` to dependency array?

dogmar (approved):

www/src/components/shell/Shell.tsx
michaeljguarino: I actually think that should be removed entirely.  Kind of surprised it didn't barf at there being no variable named ref in scope

PR #792, opened 2022-11-24
feat(frontend): Use repository name in the URL instead of id

floreks:
## Summary

Ref: [ENG-806](https://linear.app/pluralsh/issue/ENG-806/use-repository-name-instead-of-id-for-the-url-access)

- Used repository name instead of id where possible
- Updated publisher logo display to display initials when icon is not available
- Fixed gear icon misalignment on installed repository view

In order to update both publisher and chart views we would first need to update API to accept name, not only ID. @michaeljguarino do they also use unique names?

## Test Plan

Locally

## Checklist

### DIFF
--- a/www/src/components/publisher/PublisherSideNav.tsx
+++ b/www/src/components/publisher/PublisherSideNav.tsx
@@ -3,8 +3,6 @@ import { capitalize } from 'lodash'
 import { PageCard } from '@pluralsh/design-system'
 
 function PublisherSideNav({ publisher }: any) {
-  console.log(publisher)
-
   return (
     <Div
       marginLeft="medium"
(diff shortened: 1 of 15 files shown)


www/src/components/publisher/PublisherSideNav.tsx
```diff
@@ -12,6 +12,7 @@ function PublisherSideNav({ publisher }: any) {
     >
       <PageCard
         icon={{
+          name: capitalize(publisher.name),
```
michaeljguarino, line 15: are we sure we should capitalize that?  I feel like just putting the raw string is correct.

michaeljguarino (approved):
Names are guaranteed unique, so there should be no risk to this change, and will be nice to clean up our urls.  Only one small nit.

www/src/components/publisher/PublisherSideNav.tsx
floreks: I think you are right. `PageCard` takes the name and creates initials. The upper case transformation is probably already handled there.

PR #794, opened 2022-11-26
Add shell setup api

michaeljguarino:
## Summary

This will make it easier for us to surface errors to users synchronously instead of working with websocket responses

Also limits the shell update api.

## Test Plan
unit tests

## Checklist


apps/core/lib/core/services/shell/client.ex
```diff
@@ -54,4 +52,10 @@ defmodule Core.Shell.Client do
       as -> Poison.decode(body, as: as)
     end
   end
+
+  defp handle_error({:ok, %HTTPoison.Response{body: body}}) do
+    Logger.error "shell setup error: #{body}"
```
DavidSpek, line 57: I think this will probably leak credentials in plain text in the logs.

floreks: If I understand correctly this should only forward error messages returned by our `v1/setup` endpoint from the CLI. I don't think it contains any plain text credentials. This error is anyway already logged on the other side when the endpoint is called.

floreks (approved):

PR #812, opened 2022-12-02
fix(backend): module ssh_file missing during demo project

DavidSpek:
## Summary
When creating a new demo project the error below would occur. This change hopefully fixes that. We should probably add some tests for this as well.

image

## Checklist


michaeljguarino (approved):
I really doubt this will work but we can try

michaeljguarino:
The ssh key generation is tested, the problem is apparently the release generated in docker strips it out, I believe.  I'm not sure we have a good solution for testing that part until it goes live

PR #814, opened 2022-12-02
chore(artifact): use updated busybox image for migration

DavidSpek:
## Summary
With the updated image vendoring, the naming of the standard images has moved under `library/`. This PR updates the busybox image used to wait for Postgres migrations so it uses the latest image version.

## Checklist


michaeljguarino:
Looks like we need to resolve a merge conflict, but also was this tested?

DavidSpek:
Merge conflict resolved. I didn't test it explicitly, but since the latest image of busybox was being used now I doubt it could cause a problem (especially since latest is generally a tag that is changed). The image does definitely exist though so I don't see how it could break things.

PR #838, opened 2023-01-04
feat: Allow both UUIDs and names for repository URLs

dherault:
## Summary

The title is pretty self-explanatory. This would allow both these URLs to be valid:
- /repository/airbyte
- /repository/5866d30f-5cb9-47b6-9e40-535fb09f61f2

## Checklist

### DIFF
--- a/www/src/components/repository/Repository.tsx
+++ b/www/src/components/repository/Repository.tsx
@@ -8,6 +8,7 @@ import {
 } from 'react-router-dom'
 import { Flex, P } from 'honorable'
 import { Button, TabPanel } from '@pluralsh/design-system'
+import { validate as uuidValidate } from 'uuid'
 
 import { GoBack } from '../utils/GoBack'
 
@@ -28,19 +29,10 @@ import { RepositorySideCar } from './RepositorySideCar'
 
 import { REPOSITORY_QUERY } from './queries'
 
-// To test whether the provided url param is a uuid or a name
-const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i
-
 function Repository() {
   const { name } = useParams()
   const [searchParams] = useSearchParams()
-  const variables: any = { name }
-
-  if (name && uuidRegex.test(name)) {
-    variables.repositoryId = name
-  }
-
-  const { data, loading } = useQuery(REPOSITORY_QUERY, { variables })
+  const { data, loading } = useQuery(REPOSITORY_QUERY, { variables: uuidValidate(name ?? '') ? { repositoryId: name } : { name } })
   const backStackName = searchParams.get('backStackName')
   const tabStateRef = useRef<any>(null)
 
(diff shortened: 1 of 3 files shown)


www/src/components/repository/Repository.tsx
```diff
@@ -23,14 +28,23 @@ import { RepositorySideCar } from './RepositorySideCar'
 
 import { REPOSITORY_QUERY } from './queries'
 
+// To test weither the provided url param is a uuid or a name
```
dogmar, line 31:
```suggestion
// To test whether the provided url param is a uuid or a name
```

www/src/components/repository/Repository.tsx
```diff
@@ -23,14 +28,23 @@ import { RepositorySideCar } from './RepositorySideCar'
...
+const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+
```
dogmar, line 33: Since we're already using the `uuid` package, consider replacing with `uuid.validate()` instead. https://www.npmjs.com/package/uuid#uuidvalidatestr

I see we've also got `isUuid(val)` in `www/src/components/shell/validation.tsx`, but it seems to be matching for a different uuid version. You could also add your new validation in the same file and rename them with appropriate versions. Possibly both using `uuid.validate()` with `uuid.version()`

dherault: Great suggestion!

www/src/components/repository/Repository.tsx
```diff
@@ -23,14 +28,23 @@ import { RepositorySideCar } from './RepositorySideCar'
...
 function Repository() {
   const { name } = useParams()
   const [searchParams] = useSearchParams()
-  const { data } = useQuery(REPOSITORY_QUERY, { variables: { name } })
+  const variables: any = { name }
+
+  if (name && uuidRegex.test(name)) {
+    variables.repositoryId = name
```
michaeljguarino, line 40: this will technically keep the `name` attribute in the object, which probably works but it'd be safer to just do something like:

```js
variables = uuidRegex.test(name) ? { repositoryId: name} : { name}
```

(that way it doesn't rely on the precedence the api imposes to behave correctly)

www/src/components/repository/Repository.tsx
```diff
@@ -23,14 +28,23 @@ import { RepositorySideCar } from './RepositorySideCar'
...
+  }
+
+  const { data, loading } = useQuery(REPOSITORY_QUERY, { variables })
   const backStackName = searchParams.get('backStackName')
   const tabStateRef = useRef<any>(null)
 
-  if (!data) {
+  if (loading) {
```
michaeljguarino, line 47: this can make the ui flash a ton i think

dherault: No it does not, loading and data are updated in a batch

michaeljguarino: yea but it can mark as loading when the query is refetched for all sorts of reasons (new props sometimes, updates, etc)

michaeljguarino: I am pretty sure it's done it in the past though, and I don't think we should revert it because it has caused issues already unless we're sure it won't flash around

michaeljguarino (approved):

PR #850, opened 2023-01-06
WIP: resolve hydra issues

DavidSpek:
Work in progress.
## Summary
It looks like Hydra responses outside the range of `200...299` aren't being handle, which is possibly causes issues with Plural OIDC. This PR is an attempt to handle those responses.

## Checklist

### DIFF
--- a/apps/core/lib/core/clients/hydra.ex
+++ b/apps/core/lib/core/clients/hydra.ex
@@ -127,6 +127,8 @@ defmodule Core.Clients.Hydra do
 
   defp handle_response({:ok, %{status_code: code, body: body}}, type) when code in 200..299,
     do: {:ok, Poison.decode!(body, as: type)}
+  defp handle_response({:ok, %{status_code: code, body: body}}, _type) when code in 400..499,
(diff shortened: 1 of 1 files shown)


apps/core/lib/core/clients/hydra.ex
```diff
@@ -127,6 +127,8 @@ defmodule Core.Clients.Hydra do
 
   defp handle_response({:ok, %{status_code: code, body: body}}, type) when code in 200..299,
     do: {:ok, Poison.decode!(body, as: type)}
+  defp handle_response({:ok, %{status_code: code, body: body}}, _type) when code in 400..499,
```
DavidSpek, line 130: For `accept_login`, if the response is `410` it will contain a `redirect_to` which should probably be returned by our API so the user is redirected properly.

michaeljguarino: this code does handle the 409 response already, it'll fall back to the clause below and return an `{:error, :unauthorized}` tuple which will bubble to a gql error

michaeljguarino: It'll also log the error (why we're seeing them in the logs in all likelihood)

PR #851, opened 2023-01-06
Prevent double-send of accept login

michaeljguarino:
## Summary

I think this useEffect can get duplicated occasionally.  Just force it to behave well with another state var.

## Test Plan
this has to be tested in prod unfortunately

## Checklist


DavidSpek (approved):
This fixes the `AcceptLogin` mutation from running twice which is breaking Plural OIDC logins on Chrome browsers.

PR #856, opened 2023-01-11
feat: Billing 

dherault:
Adds the billing pages to the app

image
image

## Checklist

---
  - https://app.asana.com/0/0/1203932285056725


www/src/components/account/billing/BillingConsumptionProvider.tsx
```diff
@@ -0,0 +1,41 @@
...
+type BillingConsumptionProviderPropsType = {
+  children: ReactNode
+}
+
+function BillingConsumptionProvider({ children }: BillingConsumptionProviderPropsType) {
+  const { me } = useContext(CurrentUserContext)
+  const { data, loading, error } = useQuery(USERS_QUERY)
+
+  const nClusters = useMemo(() => (me?.account?.clusterCount ? parseInt(me?.account?.clusterCount) : 0), [me])
+  const nUsers = useMemo(() => data?.users?.edges?.length ?? 0, [data])
```
michaeljguarino, line 21: we should **not** use this to count users, there's a user count on the account object as well.

Also it might be more consistent to fetch the account explicitly rather than sideload it via the user since that's technically cached server side.

www/src/components/account/billing/BillingUpgradeToProfessionalModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+type BillingUpgradeToProfessionalModalPropsType = {
+  open: boolean
+  onClose: () => void
+}
+
+function BillingUpgradeToProfessionalModal({ open, onClose }: BillingUpgradeToProfessionalModalPropsType) {
+  const { proPlatformPlan } = useContext(PlatformPlansContext)
+  const { nClusters, nUsers } = useContext(BillingConsumptionContext)
+  const { card } = useContext(BillingBankCardContext)
```
michaeljguarino, line 32: You should be able to determine this w/o fetching the card. The definitive mechanism is if the account has a non null `customerId` attribute

www/src/components/account/billing/BillingUpgradeToProfessionalModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { refetch: refetchSubscription } = useContext(SubscriptionContext)
+
+  const [upgradeMutation, { loading: loadingUpgradeMutation }] = useMutation(UPGRADE_TO_PROFESSIONAL_PLAN_MUTATION, {
+    variables: {
+      attributes: {
+        lineItems: [
```
michaeljguarino, line 38: I'm going to set these server side because the more i think of it, the caching issues here are pretty gnarly

www/src/components/account/billing/BillingUpgradeToProfessionalModal.tsx
dherault: I'm fetching the card simply to display the card on the payment modal, not for the mutation.

www/src/components/account/billing/BillingSubscriptionProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+function BillingSubscriptionProvider({ children }: BillingSubscriptionProviderPropsType) {
+  const {
+    data,
+    loading,
+    error,
+    refetch,
+  } = useQuery(SUBSCRIPTION_QUERY, { fetchPolicy: 'network-only' })
+  const { proPlatformPlan, proYearlyPlatformPlan, enterprisePlatformPlan } = useContext(PlatformPlansContext)
+
+  const pricingFeaturesEnabled = useMemo(() => posthog.isFeatureEnabled('pricing'), [])
```
floreks, line 27: Did we switch from growthbook to posthog for feature flagging?

DavidSpek: Yep, we are testing that now.

PR #863, opened 2023-01-23
feat: Expose update shell mutation

floreks:
## Summary

Is there anything more I should check in the unit test? Since sensitive data is not exposed, I can't actually validate that it was updated.

## Test Plan

Locally

## Checklist

### DIFF
--- a/apps/graphql/test/mutations/shell_mutations_test.exs
+++ b/apps/graphql/test/mutations/shell_mutations_test.exs
@@ -52,6 +52,86 @@ defmodule GraphQl.ShellMutationsTest do
     end
   end
 
+  describe "updateShell" do
+    test "it will update an existing shell instance" do
+      %{email: e} = user = insert(:user, roles: %{admin: true})
+
+      expect(Pods, :fetch, fn _ -> {:ok, Pods.pod("plrl-shell-2", e)} end)
+      expect(Core.Shell.Scm, :setup_repository, fn
+        :github, ^e, "tok", nil, "demo" ->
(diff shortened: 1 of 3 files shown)


apps/graphql/test/mutations/shell_mutations_test.exs
```diff
@@ -52,6 +52,86 @@ defmodule GraphQl.ShellMutationsTest do
...
+      attrs = %{
+        "provider" => "GCP",
+        "scm" => %{"provider" => "GITHUB", "token" => "tok", "name" => "demo"},
+        "workspace" => %{
+          "cluster" => "plural",
+          "region" => "us-east-2",
+          "bucketPrefix" => "plrl",
+          "subdomain" => "demo.onplural.sh"
+        },
+        "credentials" => %{
```
michaeljguarino, line 113: fwiw, this is the only block that's honored by the update

michaeljguarino (approved):

apps/graphql/test/mutations/shell_mutations_test.exs
michaeljguarino: I probably would have written this test by using the `insert(:cloud_shell)` helper like in some other tests, but this is fine too

floreks: > fwiw, this is the only block that's honored by the update

oh yeah, I was wondering if this is what this `update_changeset` does.
```
  def update_changeset(model, attrs \\ %{}) do
    model
    |> cast(attrs, [])
    |> cast_embed(:credentials)
  end
```

PR #866, opened 2023-01-24
feat(frontend): add posthog integration and cookiebot

DavidSpek:
WIP
## Summary
This PR is an initial setup of PostHog for the frontend. We still need a cookie banner and a way for users to opt-out of tracking. We'll probably also want to add some custom events here or there, along with expanding the user properties we set.

Todo:
- [x] Add a cookie banner and allow users to opt out of tracking
- [x] Add more properties to users instead of only their email?

## Test Plan
Run locally and check PostHog to see if users are being identified and events are being recorded as expected.

## Checklist

### DIFF
--- a/www/src/components/profile/Profile.tsx
+++ b/www/src/components/profile/Profile.tsx
@@ -11,6 +11,7 @@ import {
 } from '@pluralsh/design-system'
 import { useContext, useEffect, useState } from 'react'
 import { useFilePicker } from 'react-sage'
+import posthog from 'posthog-js'
 
 import CurrentUserContext from '../../contexts/CurrentUserContext'
 import { UPDATE_USER } from '../users/queries'
@@ -44,6 +45,10 @@ export function Profile() {
   const [avatarFile, setAvatarFile] = useState<any>()
   const [mutation, { loading }] = useMutation(UPDATE_USER, {
     variables: { attributes: { name, email, avatar: avatarFile } },
+    onCompleted: ({ updateUser: { id, email } }) => {
+      posthog.identify(id)
+      posthog.people.set({ email })
+    },
   })
 
   let url = ProviderIcons[me.provider] || DEFAULT_CHART_ICON
--- a/www/src/components/profile/Profile.tsx
+++ b/www/src/components/profile/Profile.tsx
@@ -11,7 +11,6 @@ import {
 } from '@pluralsh/design-system'
 import { useContext, useEffect, useState } from 'react'
 import { useFilePicker } from 'react-sage'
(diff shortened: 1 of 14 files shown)


DavidSpek:
Thanks for the note. I'm actually not seeing the cookie banner pop up at all so that's a bit strange.

www/src/components/profile/Profile.tsx
```diff
@@ -44,6 +45,10 @@ export function Profile() {
   const [avatarFile, setAvatarFile] = useState<any>()
   const [mutation, { loading }] = useMutation(UPDATE_USER, {
     variables: { attributes: { name, email, avatar: avatarFile } },
+    onCompleted: ({ updateUser: { id, email } }) => {
```
michaeljguarino, line 48: I'm pretty sure this is unnecessary, it'll be plumbed into the use context after update

michaeljguarino (commented):
I'm not sure we need the setup in the login flows but i suppose it's not doing any damage.

We should also be positive these don't throw if posthog is down, otherwise it becomes a critical dependency on our site working.

www/src/components/profile/Profile.tsx
DavidSpek: Yeah that is still from before I added it in the `src/components/Plural.tsx`.

DavidSpek:
@michaeljguarino I just check if breaking the posthog domain by adding a random path or changing the subdomain will cause issues and it doesn't from what I can tell. There's just a bunch of failed requests in the network tab and console but the app functionality is still fine including logging in.

www/src/components/users/MagicLogin.tsx
```diff
@@ -570,6 +570,7 @@ export function Signup() {
       history.navigate('/')
     }
   }, [history])
+  // we should probably move hubspot to loaded similarly to with the other analytic tools and setup with cookiebot
   useScript({ src: 'https://js.hs-scripts.com/22363579.js' })
```
dogmar, line 574: It was originally requested that we only needed hubspot on the login page, so I only put it here to reduce the number of places we'd need to show its consent banner, but now that we're adding a global consent banner, may as well just put it up with the rest as a regular script tag marked with `type="text/plain" data-cookieconsent="statistics"`. Also will want to add some variation on [this](https://github.com/pluralsh/webflow-helpers/blob/main/standalone/tracking-mgmt.js) to disable/re-enable hubspot when prefs change.

dogmar (changes requested):

dogmar (approved):

PR #869, opened 2023-01-25
feat: Expose selected region via shell configuration workspace

floreks:
## Summary

Since `updateShell` mutation will only allow to update `credentials` struct, we will need to expose the region that can be shown in the UI. It will not be editable.

Ref: [figma](https://www.figma.com/file/985LCDEqp4ckoMLOHSWkDF/Onboarding-v3?node-id=209%3A51675&t=ZtoQFtURUFRGYXLM-4)
image

## Test Plan

Updated unit test

## Checklist

### DIFF
--- a/apps/core/lib/core/services/shell/models.ex
+++ b/apps/core/lib/core/services/shell/models.ex
@@ -7,7 +7,7 @@ defmodule Core.Shell.Models.Network do
 end
 
 defmodule Core.Shell.Models.Workspace do
-  defstruct [:network, :bucket_prefix, :cluster]
+  defstruct [:network, :bucket_prefix, :cluster, :region]
 
   def as() do
     %__MODULE__{network: %Core.Shell.Models.Network{}}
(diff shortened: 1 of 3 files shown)


michaeljguarino (approved):
This is returned in the cli api request right?  We might need another pr if not

michaeljguarino:
You can also get this from the cloud shell object itself btw (I think).  We probably shouldn't make it editable in general since it could really screw with stuff.

floreks:
Unfortunately, neither shell nor shell configuration queries expose this information right now.

michaeljguarino:
> Unfortunately, neither shell nor shell configuration queries expose this information right now.

ah we kept it out of the graphql response? We can expose it on the shell query i think, it's not sensitive info (other stuff is hidden for obv reasons)

floreks:
> > Unfortunately, neither shell nor shell configuration queries expose this information right now.
> 
> ah we kept it out of the graphql response? We can expose it on the shell query i think, it's not sensitive info (other stuff is hidden for obv reasons)

Exactly, it was defined in the model, but not exposed via shell query response. It should be safe to expose it as read-only information. I agree that we should definitely not allow to change it.

PR #878, opened 2023-01-26
feat: Wire region field through the shell query and revert shell config change

floreks:
## Summary

Another attempt at exposing `region` field :slightly_smiling_face: 

## Test Plan

Updated unit test

## Checklist

### DIFF
--- a/apps/graphql/lib/graphql/schema/shell.ex
+++ b/apps/graphql/lib/graphql/schema/shell.ex
@@ -83,6 +83,10 @@ defmodule GraphQl.Schema.Shell do
       shell, _, _ -> Shell.status(shell)
     end
 
+    field :region, non_null(:string), resolve: fn
+      %{workspace: %{region: region}}, _, _ -> {:ok, region}
+    end
+
     timestamps()
   end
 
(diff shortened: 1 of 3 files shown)


michaeljguarino (approved):

apps/graphql/lib/graphql/schema/shell.ex
```diff
@@ -98,7 +102,6 @@ defmodule GraphQl.Schema.Shell do
     field :network,       :network_configuration
     field :bucket_prefix, :string
     field :cluster,       :string
-    field :region,        :string
```
michaeljguarino, line 101: what what is this?

floreks: I just reverted this PR: https://github.com/pluralsh/plural/pull/869

Since the region is anyway null here.

floreks: oh wait, it is actually needed there for the field resolve to work, right?

michaeljguarino: if it's queried then yea it'll break if removed, I think we should probably just leave it for now

michaeljguarino (approved):

PR #894, opened 2023-02-02
feat(frontend): use type safe generated me query

DavidSpek:
## Summary
We have the `yarn graphql:codegen` command that generates types for our API, and also generates type safe Apollo query and mutation function using the queries defined in the `src/graph/*.graphql` files. This PR moves all instances of the `useQuery(ME_Q)` to the generated type safe `useMeQuery()` function. Where needed changes were made to overcome type errors. This should help prevent possible issues that weren't caught now as the `any` type was being used, as well as better autocompletion in IDEs.

The same is done for the `UPDATE_ACCOUNT` mutation that's now moved to the `useUpdateAccountMutation()` function as this had overlap with the `ME_Q` function.

This PR also updates the script to generate the TypeScript types and query/mutation functions to use a local file instead of the live version of the API. This way, changes to the API can be tested for compatibility with the frontend. It also allows for incompatible API changes can be made alongside the necessary frontend changes (in the same PR) without needing to deploy the backend.
To facilitate this, a CI test is added to the backend that will fail if the schema needs to be updated, prompting the user to run `make update-schema` and commit the changes so that the frontend can be tested with the updated API.

## Checklist


www/src/components/profile/MyProfile.tsx
```diff
@@ -26,18 +26,24 @@ const DIRECTORY = [
...
   const { pathname } = useLocation()
   const tabStateRef = useRef<any>(null)
   const currentTab = DIRECTORY.find(tab => pathname?.startsWith(tab.path))
 
+  let url: string | undefined
+
+  if (me.avatar != null) {
+    url = me.avatar
+  }
+
```
dogmar, line 39:
```suggestion
  let url = me.avatar || undefined
```
this will catch the empty string as well as being terser

dogmar (approved):

dogmar (approved):

PR #908, opened 2023-02-07
Implement new auth landing pages

dogmar:
## Summary

New designs for login/signup pages implemented.
- Updated all queries to graphql-codegen typescript versions
- Updated hubspot script to use Cookiebot for turning on/off instead of built-in HS banner
- Better messaging on invite page when user already has account

---
  - https://app.asana.com/0/0/1203879274753565


michaeljguarino:
How well-tested is this?  We definitely don't want to break auth and i know this code was complex

dogmar:
I tried each route through I could think of a few times each, but I agree this was fairly complex, so I think having more people stress it this would be a good idea.

www/src/components/users/LoginFooter.tsx
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <Div
+      flexGrow={1}
+      flexShrink={0}
+      {...props}
+    >
+      <FooterList>
+        <FooterLink as="div">
+          © Plural Labs{' '}
+          {new Date().toLocaleString('en-us', { year: 'numeric' })}
```
dherault, line 33: That I like.

dherault (approved):
👍

www/src/components/users/LoginFooter.tsx
dherault: Well, maybe my English failed me on this one. Why not getFullYear though?

dogmar: Oh, you know, just in case we ever decide to fully localize so Koreans will see `2023년` by just changing the locale string!

dogmar: #overengineering

www/src/components/users/MagicLogin.tsx
```diff
@@ -348,12 +208,11 @@ function OAuthOptions({ oauthUrls }: any) {
             wrap="wrap"
             flexWrap="wrap"
           >
-            {oauthUrls.map(url => (
+            {[...oauthUrls].sort(sortOauthUrls).map(url => (
```
michaeljguarino, line 211: why do you need to create the newlist here?

www/src/components/users/MagicLogin.tsx
```diff
@@ -363,51 +222,156 @@ function OAuthOptions({ oauthUrls }: any) {
...
+    = loginMError?.message === 'invalid password' ? 'Invalid password' : undefined
+  const loginError = !passwordErrorMsg && loginMError
 
-  const [mutation, { loading: mLoading, error }] = useMutation(LOGIN_MUTATION, {
-    variables: { email, password, deviceToken },
-    onCompleted: ({ login: { jwt } }) => {
-      setToken(jwt)
-      if (deviceToken) finishedDeviceLogin()
-      if (challenge) {
-        handleOauthChallenge(client, challenge)
```
michaeljguarino, line 392: do we still handle oauth here?

maciaszczykm (approved):

www/src/components/users/MagicLogin.tsx
dogmar: That mutation just got moved up to line 285

www/src/components/users/MagicLogin.tsx
dogmar: I didn't write that, but it does prevents sort() from mutation oauthUrls, which is not strictly necessary in this case, but I think keeping things immutable is generally good practice.

www/src/components/users/MagicLogin.tsx
michaeljguarino: ahh

michaeljguarino (approved):

PR #913, opened 2023-02-08
feat: Send custom events to posthog on onboarding/installer error

floreks:
## Summary

Since we can't catch any errors from inside the terminal, I have captured errors from the installer and during shell creation.

I have added 2 new custom event types that will be visible in posthog:
- `Installer`
- `Onboarding`

### Onboarding
Consists of:
- error object
- provider
- cluster name
Zrzut ekranu z 2023-02-08 15-20-13
Zrzut ekranu z 2023-02-08 15-19-40

### Installer
- error object
- application list (names only)
- provider
Zrzut ekranu z 2023-02-08 15-48-07
Zrzut ekranu z 2023-02-08 15-48-21

## Test Plan

Locally + posthog

## Checklist

### DIFF
--- a/www/src/utils/posthog.ts
+++ b/www/src/utils/posthog.ts
@@ -1,10 +1,14 @@
 import posthog from 'posthog-js'
 
-import { User } from '../generated/graphql'
+import { ApolloError } from '@apollo/client'
+
+import { Provider, User } from '../generated/graphql'
+
+import { CloudProvider } from '../components/shell/onboarding/context/types'
 
 import Cookiebot from './cookiebot'
 
-export default function PosthogIdentiy(me: User) {
+export default function PosthogIdentify(me: User) {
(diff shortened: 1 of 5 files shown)


www/src/utils/posthog.ts
```diff
@@ -1,10 +1,14 @@
...
+import { ApolloError } from '@apollo/client'
+
+import { Provider, User } from '../generated/graphql'
+
+import { CloudProvider } from '../components/shell/onboarding/context/types'
 
 import Cookiebot from './cookiebot'
 
-export default function PosthogIdentiy(me: User) {
+export default function PosthogIdentify(me: User) {
```
DavidSpek, line 11: This change will also need to be propagated to where this function is used

DavidSpek (approved):
LGTM

PR #929, opened 2023-02-09
feat: Add attributes to various elements for posthog targeting

dogmar:
## Summary

Added some `data-phid` attributes to help target things in posthog

## Checklist

### DIFF
--- a/www/src/components/layout/Subheader.tsx
+++ b/www/src/components/layout/Subheader.tsx
@@ -83,7 +83,7 @@ export default function Subheader() {
           type="floating"
         />
         <IconFrame
-          data-phid="nav-arrow-fwd"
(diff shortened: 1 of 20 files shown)


www/src/components/layout/Subheader.tsx
```diff
@@ -82,6 +83,7 @@ export default function Subheader() {
           type="floating"
         />
         <IconFrame
+          data-phid="nav-arrow-fwd"
```
DavidSpek, line 86: I don’t think we’re using abbreviations anywhere else so to keep it consistent I’d use `nav-arrow-forward`.

DavidSpek (approved):
Overall LGTM from reading through it on my phone.

PR #941, opened 2023-02-11
fix: Notifications are hiding new app nav bar

dogmar:
## Summary
- New context/hook for turning on/off overlay of main content area only
- Added in/out transitions for Notifications panel
- Trying out a blur effect for the overlay. I like it, but it's not consistent with our Modals. Maybe we bring it to our modals too?

https://user-images.githubusercontent.com/85062/218228648-476b9316-00d7-446e-86cc-e33f1b52a8ba.mov

## Checklist

---
  - https://app.asana.com/0/0/1203878978472133


maciaszczykm (approved):
Blur looks great, would be really nice to add it for modals as well in my opinion.

dogmar:
I'm going to remove the blur effect for now, until there's a consensus on adding it across the whole app.

PR #954, opened 2023-02-17
Add billing info on useBankCard

dherault:
Adds the billing address to useBankCard to have it alongside any card form.
image

## Checklist

### DIFF
--- a/www/src/components/account/billing/useBankCard.tsx
+++ b/www/src/components/account/billing/useBankCard.tsx
@@ -7,23 +7,34 @@ import {
   useState,
 } from 'react'
 import { useMutation } from '@apollo/client'
-import { Button, Card } from '@pluralsh/design-system'
+import {
+  Button,
+  Card,
+  FormField,
+  Input,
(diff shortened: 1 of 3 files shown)


www/src/components/account/billing/useBankCard.tsx
```diff
@@ -84,13 +125,109 @@ function useBankCard(setEdit: Dispatch<SetStateAction<boolean>>, address, noCanc
     setLoading(false)
   }, [card, deleteCard, refetch])
 
+  const renderBillingForm = useCallback(() => (
```
michaeljguarino, line 128: this pattern is very bug prone and causes gigantic run-on functions.  If we could split it into an entirely separate function component or another file, that would probably be ideal for readability

michaeljguarino (commented):
This shouldn't be required to set on each card that's input, only the first one if it was never set.  After a billing address is persisted, we should probably have a separate tab w/in accounts to update billing info as well but separate from here.

PR #961, opened 2023-02-21
Posthog user.created event, and wipe root user id when moving accounts

michaeljguarino:
## Summary
We've seen a few users become undeletable because they organically signed up pre-invite and had an account record still point to them as root user.  This should fix that and backfill some of the stale data.

## Test Plan
unit tests

## Checklist

### DIFF
--- a/apps/core/lib/core/pubsub/protocols/hoggable.ex
+++ b/apps/core/lib/core/pubsub/protocols/hoggable.ex
@@ -19,3 +19,7 @@ defimpl Core.PubSub.Hoggable, for: [Core.PubSub.InstallationCreated, Core.PubSub
   defp event(PubSub.InstallationCreated), do: "installation.created"
   defp event(PubSub.InstallationDeleted), do: "installation.deleted"
 end
+
+defimpl Core.PubSub.Hoggable, for: Core.PubSub.UserCreated do
+  def hog(%{item: user}), do: {"user.created", user.id, %{email: user.email}}
+end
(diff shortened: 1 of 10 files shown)


apps/core/lib/core/pubsub/protocols/hoggable.ex
```diff
@@ -19,3 +19,7 @@ defimpl Core.PubSub.Hoggable, for: [Core.PubSub.InstallationCreated, Core.PubSub
   defp event(PubSub.InstallationCreated), do: "installation.created"
   defp event(PubSub.InstallationDeleted), do: "installation.deleted"
 end
+
+defimpl Core.PubSub.Hoggable, for: Core.PubSub.UserCreated do
+  def hog(%{item: user}), do: {"user.created", user.id, %{email: user.email}}
```
DavidSpek, line 24: We can’t set the user email here since we’d need to check consent for that (which is done with a cookie on the frontend). But this shouldn’t be a problem since if the user accepts cookies their email will get associated with this ID in PostHog later on.

DavidSpek (changes requested):
We shouldn’t set the user’s email in the event since we don’t have a way to track consent for that in the backend.

DavidSpek:
Also, can you run `make update-schema` and push those changes.

apps/core/lib/core/pubsub/protocols/hoggable.ex
michaeljguarino: Shouldn't that become null if they sign up though (I'm not sure you can even sign up w/o cookies actually)

DavidSpek: You can signup without cookie tracking. It won’t become null, since we use the Plural UUID to track users. So if they never consent to cookies we would have marketing site visits of a user that aren’t associated with this signup event. However, as soon as they consent all the anonymous PostHog IDs will get associated with the Plural UUID and it will set the user’s email on their person in PostHog.

DavidSpek (approved):
LGTM

PR #969, opened 2023-02-27
Implement cluster promotions

michaeljguarino:
## Summary
The way this is going to work is a user has a timestamp if promotions are enabled that forces upgrades to be deferred up to that timestamp, and each promotion will simply update that waterline.

Also cleaned up some of the upgrade code overall.  I've deferred gating it on the multi_cluster feature since we might want to let some users test it early-access

## Test Plan
unit tests throughout

## Checklist


DavidSpek (approved):
LGTM to unblock

PR #974, opened 2023-03-02
Expose demoed field in the GQL User object

floreks:
## Summary

This will be useful in the onboarding to disable demo option if user has reached demo limit.

## Test Plan

Added unit test

## Checklist

### DIFF
--- a/apps/core/lib/core/services/shell/demo.ex
+++ b/apps/core/lib/core/services/shell/demo.ex
@@ -25,6 +25,9 @@ defmodule Core.Services.Shell.Demo do
   @lock "demo-projects"
   @max_count 3
 
+  @spec max_demo_count :: binary
+  def max_demo_count, do: @max_count
+
   @spec has_demo?(binary) :: boolean
   def has_demo?(user_id) do
     DemoProject.for_user(user_id)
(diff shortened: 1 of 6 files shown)


apps/core/lib/core/services/shell/demo.ex
```diff
@@ -25,6 +25,9 @@ defmodule Core.Services.Shell.Demo do
   @lock "demo-projects"
   @max_count 3
 
+  @spec max_demo_count :: binary
```
floreks, line 28: should it be `integer` instead of `binary`?

apps/graphql/lib/graphql/schema/user.ex
```diff
@@ -132,6 +132,10 @@ defmodule GraphQl.Schema.User do
       user, _, _ -> {:ok, User.background_color(user)}
     end
 
+    field :demoed, :boolean, resolve: fn
```
DavidSpek, line 135: Can you add documentation here. You can see https://github.com/pluralsh/plural/blob/master/apps/graphql/lib/graphql/schema/repository.ex#L196 for an example.

apps/core/lib/core/services/shell/demo.ex
DavidSpek: I think it would be an integer.

michaeljguarino (approved):

PR #1004, opened 2023-03-10
fix(frontend): add address element and fix add cards

DavidSpek:
## Summary
Our address input not doesn't have any form of validation and is completely missing when users want to add a card that isn't in the `Upgrade now` flow. This PR adds the stripe `AddressElement` which changes the fields based on the country a user has selected, has proper validation and can also do things like autocomplete.

The `AddressElement` can be styled by changing using the `appearance` options set on the `Elements` components. The docs for the appearance API can be found [here](https://stripe.com/docs/elements/appearance-api). The docs for getting the information from the AddressElement and further validating the inputs are [here](https://stripe.com/docs/elements/address-element/collect-addresses).

**Please change the hostname in `www/src/helpers/hostname.ts` to `app.dev.plural.sh` for any testing.**

## Checklist

### DIFF
--- a/www/src/components/Plural.tsx
+++ b/www/src/components/Plural.tsx
@@ -121,6 +121,16 @@ function WrapStripe({ children }: any) {
     appearance: {
       theme: 'night',
       labels: 'floating',
+      variables: {
+        // colorPrimary: '#747B8B',
+        // colorBackground: '#747B8B',
+        fontSizeBase: '10px',
+        colorText: '#747B8B',
+        // colorDanger: '#df1b41',
+        // fontFamily: 'Ideal Sans, system-ui, sans-serif',
+        // spacingUnit: '2px',
+        // borderRadius: '4px',
+      },
     },
   }
 
--- a/www/src/components/Plural.tsx
+++ b/www/src/components/Plural.tsx
@@ -117,10 +117,20 @@ function WrapStripe({ children }: any) {
 
   const stripePromise = useMemo(() => loadStripe(stripePublishableKey), [stripePublishableKey])
 
+  const options = {
+    appearance: {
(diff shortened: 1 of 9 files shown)


swoodward90 (commented):
One thing I think we need to address - for whatever reason the text on these form fields is rendering very dark for me to the point it's difficult to see. I think we need to figure out how to lighten that text up:
Screenshot 2023-03-10 at 1 50 41 PM

We should also remove the form fields we added in favor of this one so we're not having folks enter their information twice

DavidSpek:
The text and background colors can luckily all easily be changed, along with the spacing between the inputs. It mainly needs to be integrated into the surrounding components nicely. Along with the actual functionality.

DavidSpek:
Some things that still need fixing:
- [ ] Validation is missing that the card issuing country is the same as the billing address country
- [ ] The address input for validation can fail after a reload. Leaving the user unable to continue easily
- [ ] The `Add card` button is clickable when the for validation has failed
- [ ] It's possible that an old billing address is saved somewhere in Stripe or our API and when adding a new card the billing address is the old address and not the one entered when adding the new card.
- [ ] Validation that the card input is correct. For American Express cards the CVC is 4 digits but it will accept 3 on input.

swoodward90:
@DavidSpek / @dogmar looks like there's still the issue with an old Billing Address being retained even if that card is deleted and a new one is entered. Any idea what's going on here?

DavidSpek:
@swoodward90 Im not sure if that is coming from our API saving the address somewhere, it not being updated in Stripe properly, our API not getting the new address from Stripe or something else.

dogmar (approved):

swoodward90 (approved):

PR #1012, opened 2023-03-16
Add support for payment intents

michaeljguarino:
## Summary
This pr does two things:
* allow on-the-fly stripe customer creation when creating subscriptions
* allow querying latest invoice for a subscription to fetch payment intent+client secret

This should let us support the full payment intent flow to see if it's more robust in EU

## Test Plan
unit tests

## Checklist


swoodward90:
ok quick thoughts - 
- do we want to add the ability to use SetupIntent here too (when adding a card?) - that seems like the proper way to do it without creating a subscription at that time
-  should we be attaching billing address to a PaymentMethod instead of an account?

www/src/graph/payments.graphql
```diff
@@ -53,6 +71,9 @@ fragment Invoice on Invoice {
   createdAt
   hostedInvoiceUrl
   lines { ...InvoiceItem }
+  paymentIntent {
```
michaeljguarino, line 74: This shouldn't be on all invoice structs, since it won't be sideloaded on listing invoices for an account for example

maciaszczykm (approved):

PR #1021, opened 2023-03-21
Stripe payment intents front-end

dogmar:


www/src/components/account/billing/ConfirmPayment.tsx
```diff
@@ -0,0 +1,178 @@
...
+      try {
+        stripe
+          .retrieveSetupIntent(clientSecret)
+          .then(({ setupIntent, error }) => {
+            console.log('stripe confirmCardPayment returned, ',
+              setupIntent,
+              error)
+
+            setError(error)
+            setPaymentIntent(setupIntent)
```
michaeljguarino, line 71: we need to set this as the default payment method before upgrading so the subscription will actually bill the card

michaeljguarino: I can add this to the createSubscription api too if it'll make things simpler or we can do it in the frontend

michaeljguarino: (but generally the setup intent form should always add that payment method as the new default)

www/src/graph/payments.graphql
```diff
@@ -114,36 +136,88 @@ query Subscription {
...
+  id
+  type
+  isDefault
+  card {
+    ...Card
+  }
+}
+
+query PaymentMethods {
+  account {
```
michaeljguarino, line 204: we should ideally have just one account query for all needed data on the billing page (ie paymentMethods, subscription, etc).  You might also need to make it resistent to errors, since i think currently only those w/ billing access can query cards on file

www/src/components/account/billing/BillingPricingCards.tsx
```diff
@@ -95,23 +98,30 @@ function BillingPricingCards() {
...
+                onClick={
+                  isEnterprisePlan ? null : () => setDowngradeModalOpen(true)
+                }
+              >
+                Downgrade
+              </Button>
+            ) : isEnterprisePlan ? (
+              <ContactUs />
+            ) : (
+              renderCurrentPlanButton()
```
michaeljguarino, line 115: Could we refactor this into another function component?  These `useCallback` subcomponents are crazily bug-prone

www/src/components/account/billing/BillingBankCards.tsx
```diff
@@ -1,70 +1,243 @@
...
+    <Modal
+      open={open}
+      onClose={onClose}
+      header="Add payment method"
+      size="large"
+    >
+      <PaymentForm
+        formVariant={PaymentFormVariant.AddCard}
+        onClose={() => {
+          console.log('onClose modal')
```
michaeljguarino, line 155: rm

www/src/components/account/billing/ConfirmPayment.tsx
```diff
@@ -0,0 +1,276 @@
...
+    = useUpgradeToProfessionalPlanMutation({
+      variables: { planId },
+      refetchQueries: [namedOperations.Query.Subscription],
+      onCompleted: () => {
+        setUpgradeSuccess(true)
+      },
+    })
+
+  // Confirm payment
+  const onComplete = useCallback(() => {
```
michaeljguarino, line 176: is this useCallback necessary?

www/src/components/account/billing/ConfirmPayment.tsx
```diff
@@ -0,0 +1,276 @@
...
+  },
+  [upgradeMutation])
+  const { error: setDefaultError, setupIntent } = useConfirmAndSetDefault({
+    clientSecret,
+    onComplete,
+  })
+
+  const error = upgradeError || setDefaultError
+
+  useEffect(() => {
```
michaeljguarino, line 187: I'm pretty sure this should no longer be an issue if you want to drop that annoying error handling

www/src/components/account/billing/PaymentMethodsContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+        defaultPaymentMethod: PaymentMethodFragment | undefined
+      })
+
+    return {
+      paymentMethods: result.paymentMethods,
+      defaultPaymentMethod: result.defaultPaymentMethod,
+    }
+  }, [data?.account?.paymentMethods?.edges])
+
+  console.log('paymentMethods', paymentMethods)
```
michaeljguarino, line 61: rm

www/src/components/account/billing/ConfirmPayment.tsx
dogmar: Are you talking about the 'account_id has already been taken' error? It can still crop up if the user refreshes after the upgrade has already succeeded. Though I guess I could just check if they're already upgraded before attempting the mutation.

michaeljguarino: yea i think it should be fine, on the refresh, it should load the page w/ the subscription pinned to the pro plan so i'd assume it wouldn't cause an issue right?

PR #1032, opened 2023-03-29
feat: Payment Intents Improvements

dogmar:


www/src/components/account/billing/PaymentForm.tsx
```diff
@@ -567,36 +563,50 @@ function AddressForm({
...
     = plan === 'yearly' ? proYearlyPlatformPlan.id : proPlatformPlan.id
 
   // Upgrade mutation
-  const [upgradeMutation, { loading, error }] = useUpgradeSubscription(
-    { planId },
-    (result, nextPath) => navigate(`${nextPath}&payment_intent_client_secret=${result.paymentIntent.client_secret}`),
-    () => null,
-    () => setUpgradeSuccess(true)
-  )
+  const [upgradeMutation, { loading }] = useCreatePlatformSubscriptionMutation({
```
michaeljguarino, line 576: why not use the `useUpgradeSubscription` function here?

dogmar: This way I can just redirect to `<ConfirmPayment>` to handle both `stripe.confirmPayment` and all the error/success messaging.

michaeljguarino: Oh i think i see the difference here.

maciaszczykm (approved):

PR #1049, opened 2023-04-06
Downgrade cli to avoid recent push bug

michaeljguarino:
## Summary
new versions of the cli didn't properly implement `plural template` so need to downgrade until fixed to unblock

## Test Plan
n/a

## Checklist


DavidSpek (approved):
LGTM. Although we need to fix this in the CLI so we don't get stuck with using outdated versions of our own code.

PR #1211, opened 2023-08-23
Sideload lock state on installations

michaeljguarino:
## Summary
This can make it more visible that a manual upgrade is pending for an installation and what to do

## Test Plan
unit test

## Checklist


michaeljguarino:
That would be a lot more difficult, at the moment I'm just going to try to make it more visible and think of a solution for the rest

PR #1215, opened 2023-08-25
Feat: ai from s3

rauerhans:
## Summary

After the dagster ETL for the plural ai vector store index is validated to work in prod, this should be merged to get the data from the pipeline sink at s3. I left the old scraper as is for now, we can certainly remove it when everything works as intended.

The storage context is now pulled from s3 so the `main.py` script needs to know where to find it and how to authenticate.

- Auth:
    IRSA should work, otherwise you'll need to set the standard AWS env vars:
    - `AWS_ACCESS_KEY_ID`
    - `AWS_SECRET_ACCESS_KEY`
- Path:
    The script expects the AWS path in `PLURAL_AI_INDEX_S3_PATH` in the format `<bucket-name>/<path>`.
    Defaults to `plural-assets/dagster/plural-ai/vector_store_index`

To be safe `AWS_DEFAULT_REGION` should be set to the region of the bucket.

## Checklist


ai/main.py
```diff
@@ -22,7 +23,11 @@ class QueryResponse(BaseModel):
 service_context = ServiceContext.from_defaults(embed_model=embed_model)
 set_global_service_context(service_context)
 
-storage_context = StorageContext.from_defaults(persist_dir="./storage")
+storage_context = StorageContext.from_defaults(
```
michaeljguarino, line 26: Will this reload if the s3 data changes on a regular interval or would we need to restart the pod to get it to do that?

My general guess was we might need to have a thread that polls the bucket and reinitializes the storage context (or a sidecar that downloads it to a local file that can be read out)

rauerhans: good point

rauerhans: the quickest way to solve it is probably with a fastapi background task or sth, will look into it how this deals with global objects

rauerhans: this would do it I think, probably not super nice doing it just with asyncio, there might be some fastapi features that are more suited, but it should work for the time being

michaeljguarino (approved):

PR #1219, opened 2023-08-28
fix: Could not change automatic upgrades setting

dogmar:


www/src/components/app/upgrade/Upgrade.tsx
```diff
@@ -6,36 +6,50 @@ import {
...
+  const { autoUpgrade, trackTag } = formState
   const [mutation, { loading, error }] = useMutation(UPDATE_INSTALLATION, {
-    variables: { id: installation?.id, attributes: { trackTag, autoUpgrade } },
+    variables: {
+      id: installation?.id,
+      attributes: { trackTag, autoUpgrade },
+    },
   })
   const hasUpgradeChannels = useMemo(
     () => !isEmpty(upgradeChannels),
```
michaeljguarino, line 43: this definitely shouldn't be reading as empty most of the time, apps will at least always have a latest channel

michaeljguarino (approved):

www/src/components/app/upgrade/Upgrade.tsx
michaeljguarino: oh i see we needed to add the `upgradeChannels` field to the gql query as well

PR #1222, opened 2023-08-28
feat: Reduce load time on /marketplace

dogmar:
Remove unnecessary field requests in `MarketplaceRepo` query and `Installation` fragments


www/src/graph/installations.graphql
```diff
@@ -18,31 +29,3 @@ fragment Installation on Installation {
     ...OIDCProvider
   }
 }
-
-query GetInstallation($name: String) {
```
michaeljguarino, line 22: is it ok these are all removed?

dogmar: Yup, I verified that they weren't used anywhere.

michaeljguarino: kk

michaeljguarino (approved):

PR #1243, opened 2023-09-06
Add shell uniqueness detection

michaeljguarino:
## Summary

Some terminal session managers behave strangely when there are too many clients contending for a session, this should add relatively robust checks for that.  There are a few cases:

1.  A websocket networks to a running channel process in-cluster, in that case we should detect it already has a wss established and not connect again.
2.  A websocket creates an entirely fresh channel somewhere in-cluster and there's already a live process elsewhere, we can rpc to that node check if its alive, then bail if so.
3.  A websocket creates a fresh channel, but the adjacent process has died.  We will detect this by either seeing its node died or an rpc to a live node w/o that process fails to see the process there.

Also added a feature gate like we have for sysbox to test end-to-end

## Test Plan
some unit tests, need end to end tests to really verify though

## Checklist

### DIFF
--- a/apps/core/lib/core/schema/cloud_shell.ex
+++ b/apps/core/lib/core/schema/cloud_shell.ex
@@ -1,6 +1,6 @@
 defmodule Core.Schema.CloudShell do
   use Piazza.Ecto.Schema
-  alias Piazza.Ecto.EncryptedString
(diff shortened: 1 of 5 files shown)


DavidSpek (approved):
LGTM

PR #1248, opened 2023-09-12
chore: update CLI version used by Plural

davidspek:
## Summary
This PR updates the CLI version used by Plural to the latest release. This is needed as pushing charts with the new `chartInstalled` template function currently fail (see https://github.com/pluralsh/plural-artifacts/actions/runs/6159915713/job/16715910451).


michaeljguarino (approved):

michaeljguarino:
there was some issue a while back that required us to use that older version though and i can't remember what it was

michaeljguarino:
I think there was a missing template command but it should be fine now

DavidSpek:
It was `plural template` but I just ran that locally and it seems to be working as it should.

PR #1351, opened 2024-08-27
cloud setup flow

jsladerman:
adds ui for cloud-hosted console


floreks:
Quickly went through the screens to do some VQA based on the [figma design](https://www.figma.com/design/Kn0s2plNxcvpPrlau72z1v/%F0%9F%92%BB-Console-1.0?node-id=8309-2&node-type=CANVAS&t=35E7ulj1oc8cW9Dw-0). Is it the latest design? I am seeing some differences that seem intentional.

### Install Plural CLI
#### Current
image
#### Spec
image

### Deploy Yourself Path
I didn't see the `Your cluster was successfully deployed!` toast after completing it.
image

### Use Plural Cloud Path
- I am not seeing the free trial user notification
#### Current 
image
#### Spec
image

- Navigating out of the cloud instance creation view after it has started creating the cloud console in the background does not trigger any warning dialog. There is no way to get back and resume afterward. I imagine that API still creates it in background.
- I am not able to access the cloud console
image

floreks:
#### Zerostate for the self-hosted clusters is empty.
image

We can either:
- auto-switch to cloud instances tab if it is empty
- add a zerostate (empty table or something)

#### Trying to access `Edit cluster managers` dialog when console is in pending state fails
image

It is due to SA being undefined. Probably because instance is not yet fully provisioned.
image

#### Trying to access console now fails due to `redirectUris` not being configured for the OIDC provider.
image

It might be best to add this field, similar to the console edit oidc view. Our main `app.plural.sh` should be configured by default. I am not sure why it isn't. Maybe the mutation that updates the user, deletes previously configured `redirectUris`?

jsladerman:
> #### Zerostate for the self-hosted clusters is empty.
> image
> 
> We can either:
> 
> * auto-switch to cloud instances tab if it is empty
> * add a zerostate (empty table or something)
> 
> #### Trying to access `Edit cluster managers` dialog when console is in pending state fails
> image
> 
> It is due to SA being undefined. Probably because instance is not yet fully provisioned. image
> 
> #### Trying to access console now fails due to `redirectUris` not being configured for the OIDC provider.
> image
> 
> It might be best to add this field, similar to the console edit oidc view. Our main `app.plural.sh` should be configured by default. I am not sure why it isn't. Maybe the mutation that updates the user, deletes previously configured `redirectUris`?

addressed all these in latest commit

floreks:
I retested everything and haven't found any issues. Everything works smoothly now. Looks great!

www/src/components/overview/clusters/plural-cloud/ConsoleInstanceOIDC.tsx
```diff
@@ -0,0 +1,217 @@
...
+            : redirectUris,
+          bindings: bindings.map((value) =>
+            sanitize(value as SanitizePropsType)
+          ),
+        },
+      },
+      onCompleted: onClose,
+    })
+
+  if (!provider?.bindings || loadingRepo || errorRepo) return null
```
floreks, line 112: nit: There is this initial dialog flicker when initializing the data where it renders an empty dialog first. We can either render a normal-sized dialog and show a loading spinner or wrap the dialog in a suspense that can fallback to the loading overlay until the data is loaded.

image

jsladerman: fixed, went with the suspense option i think it looks good

michaeljguarino (approved):

PR #1404, opened 2025-03-06
add TTL to create cluster flow cached state

jsladerman:
also fixes some links to the marketing site


michaeljguarino (commented):
having a separate item store the ttl information seems intrinsically brittle, it might be better if you just wrap the object perhaps and store it under the same key.

PR #1485, opened 2026-05-27
fix: Verify reCAPTCHA on submit

maciaszczykm:
Verify reCAPTCHA on submit instead of loading the token on page load and then using it in login mutation. Doing it in two phases could lead to validation failures as tokens are short-lived. It also removes token from the state and the manual verify reCAPTCHA link under the password field.

## Test Plan
Tested locally both success path and a path where token is invalid (modified in code to check if it will fail with correct error).

## Checklist

### DIFF
--- a/www/src/components/users/MagicLogin.tsx
+++ b/www/src/components/users/MagicLogin.tsx
@@ -408,16 +408,14 @@ function LoginInternal() {
   })
 
   const isPasswordLogin =
-    state === State.PwdLogin ||
-    state === State.PwdLogin_CheckingPwd
+    state === State.PwdLogin || state === State.PwdLogin_CheckingPwd
   const disableSubmit = isPasswordLogin
     ? password.length === 0 || isVerifyingCaptcha || !executeRecaptcha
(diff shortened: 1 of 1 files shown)


maciaszczykm:
@greptileai

maciaszczykm:
@greptileai

www/src/components/users/MagicLogin.tsx
```diff
@@ -404,17 +403,42 @@ function LoginInternal() {
...
+
+    setState(State.PwdLogin_CheckingPwd)
+    loginMutation({
+      variables: {
+        email,
+        password,
+        deviceToken: typeof deviceToken === 'string' ? deviceToken : undefined,
+        captcha,
+      },
+    }).catch(() => undefined)
```
michaeljguarino, line 433: should it do this catch?

maciaszczykm: Errors will still surface via `loginMError` and without it we would get unhandled promise rejection.

michaeljguarino (approved):

